PR #3, opened 2021-08-24
WIP : Add 2D convolution benchmark using Boost.GIL

meshtag:
This PR intends to add a 2D convolution benchmark using Boost.GIL. CMake script assumes that the user has GIL and related dependencies installed in their system. 
GIL uses `libpng` as backend for reading images with `.png` extension. It is a known issue that `libpng` is not compatible with all `.png` images. To solve this problem, we can use opencv's `imread()` to read an image and then manually convert it to GIL's `gil::view` container. Is this the right approach to handle the above problem? (/cc @zhanghb97 )
I will add required piece(s) of documentation related to running the benchmark once we solve the above problem.


zhanghb97:
>  GIL uses libpng as backend for reading images with .png extension. It is a known issue that libpng is not compatible with all .png images.

Can you give me more information about this issue?

meshtag:
There are compatibility issues while reading some images with `.png` extension. For example, `gil::read_image()` was unable to read [this](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/conv-opt/images/YuTu.png) image on my system.

zhanghb97:
Will GIL fix this problem in the future? Now you can use OpenCV to read an image, I think it will not affect the performance of the convolution. When the GIL supports png, we will change it then.

examples/conv-opt/comparison/boost_gil_conv2d.cpp
```diff
@@ -0,0 +1,43 @@
+#include <boost/gil/extension/io/png.hpp>
```
zhanghb97, line 1: Please format the file with `clang-format`.

zhanghb97 (changes requested):
The GIL result image is different from the OpenCV and buddy `conv-opt` tool. Did you notice this before?

meshtag:
> Did you notice this before?

Yes, that is because `OpenCV` uses `BORDER_DEFAULT` (same as `BORDER_REFLECT_101`) as the default method of boundary extrapolation whereas this version of `GIL` assigns the value of zero to extrapolated pixels(similar to `BORDER_CONSTANT`). Also, `buddy conv-opt` has support for only one position of anchor point(which is (0, 0)) while the default position of anchor point(which is used in this case) for both `GIL` and `OpenCV` is (1, 1) for a 3x3 kernel. Above are the reasons for different outputs.

examples/conv-opt/comparison/boost_gil_conv2d.cpp
meshtag: Since I couldn't find a custom styled `.clang-format` file, I have used the default one generated using `clang-format -style=llvm -dump-config > .clang-format`.

zhanghb97:
Thanks! I have resolved the conflicts and added find_package for boost.

PR #7, opened 2021-12-02
2D Correlation documentation

meshtag:
This PR intends to add basic documentation for 2D correlation in DIP dialect.

### DIFF
--- a/docs/dip-opt.md
+++ b/docs/dip-opt.md
@@ -9,7 +9,6 @@ applications in which boundary extrapolation is not explicitly required. Due to 
 using linalg dialect's 2D correlation API.
 
 dip.corr_2d performs boundary extrapolation for making the size of output image equal to the size of input image and then uses coefficient broadcasting and strip mining(CBSM) approach for performing correlation as shown in following schematic : 
-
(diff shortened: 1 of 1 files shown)


zhanghb97 (commented):
The overall content LGTM. I have a suggestion on markdown format. In a paragraph of text, I suggest using one line for one sentence.

For example:
current
```
line 7 - This operation is used for performing 2D correlation on an image. The 2D correlation API provided by the linalg dialect is more suited for
line 8 - applications in which boundary extrapolation is not explicitly required. Due to this, dimensions of output are always less than the input dimensions after
line 9 - using linalg dialect's 2D correlation API.
```
one line for one sentence:
```
line 7 - This operation is used for performing 2D correlation on an image. 
line 8 - The 2D correlation API provided by the linalg dialect is more suited for applications in which boundary extrapolation is not explicitly required. 
line 9 - Due to this, dimensions of output are always less than the input dimensions after using linalg dialect's 2D correlation API.
```

What do you think about this?

zhanghb97:
LGTM! Please resolve the conflicts.

zhanghb97 (approved):
Thanks！

PR #13, opened 2022-02-20
Improve DIP dialect interface

meshtag:
DIP dialect specific functions/entities were defined/declared in a separate header file named `dip.hpp`. Proposed C++ interface for `dip.corr_2d` provides complete control over its functionalities to the user. A sample depicting its usage is : 
```
dip::Corr2D(input, kernel, output, centerX, centerY, boundary_option, constantValue(optional));
``` 
where 
```
input : Input image memref

kernel : Kernel memref

output : Output image memref

centerX : X co-ordinate of anchor point

centerY : Y co-ordinate of anchor point

boundary_option : Desired method of boundary extrapolation. 
Available options are dip::BOUNDARY_OPTION::CONSTANT_PADDING and dip::BOUNDARY_OPTION::REPLICATE_PADDING. 

constantValue : Optional argument which is expected to be provided with dip::BOUNDARY_OPTION::CONSTANT_PADDING. Its default value is zero.
```

Note : I did not include `Memref_descriptor` and related things in `namespace dip {}` because its implementation is not yet finalised. I will make necessary changes once `Memref_descriptor` container is successfully ported from buddy_benchmark repository.

### DIFF
--- a/examples/DIPDialect/correlation2D.cpp
+++ b/examples/DIPDialect/correlation2D.cpp
@@ -15,42 +15,10 @@
 
 #include <iostream>
 #include <time.h>
+#include <dip.hpp>
 
 using namespace cv;
 using namespace std;
-
-// Define Memref Descriptor.
-typedef struct MemRef_descriptor_ *MemRef_descriptor;
-typedef struct MemRef_descriptor_ {
-  float *allocated;
-  float *aligned;
-  intptr_t offset;
-  intptr_t sizes[2];
-  intptr_t strides[2];
-} Memref;
-
-// Constructor
-MemRef_descriptor MemRef_Descriptor(float *allocated, float *aligned,
-                                    intptr_t offset, intptr_t sizes[2],
-                                    intptr_t strides[2]) {
-  MemRef_descriptor n = (MemRef_descriptor)malloc(sizeof(*n));
-  n->allocated = allocated;
-  n->aligned = aligned;
-  n->offset = offset;
-  for (int i = 0; i < 2; i++)
-    n->sizes[i] = sizes[i];
-  for (int j = 0; j < 2; j++)
-    n->strides[j] = strides[j];
-
-  return n;
-}
-
-// Declare the Corr2D C interface.
-extern "C" {
-void _mlir_ciface_corr_2d(MemRef_descriptor input, MemRef_descriptor kernel,
-                          MemRef_descriptor output, unsigned int centerX,
-                          unsigned int centerY, int boundaryOption);
-}
 
 bool testImages(cv::Mat img1, cv::Mat img2) {
   if (img1.rows != img2.rows || img1.cols != img2.cols) {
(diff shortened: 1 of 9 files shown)


zhanghb97:
When I build this patch with the latest buddy-mlir and run the example, I find that the buddy-opt report a issue:
```
buddy-opt: for the --DIP-strip-mining option: '' value invalid for long argument!
```
And the linker cannot find the `_mlir_ciface_corr_2d_constant_padding` and `_mlir_ciface_corr_2d_replicate_padding` for the `correlation2D` example.
Please check these and see if you can reproduce it.

include/dip.hpp
```diff
@@ -0,0 +1,82 @@
+//===- dip.hpp --------------------------------------------------------===//
```
zhanghb97, line 1: I propose using `*.h` for the consistency and the directory structure like:
```
- include
    - Dialect
    - Target
    - Interface
        - buddy
            - dip.h
```
Maybe we will have a "buddy-c" directory for CAPI and potential python binding.

include/dip.hpp
```diff
@@ -0,0 +1,82 @@
...
+//
+// Header file for DIP dialect specific operations and other entities.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INCLUDE_DIP
+#define INCLUDE_DIP
+
+// Define Memref Descriptor.
+typedef struct MemRef_descriptor_ *MemRef_descriptor;
```
zhanghb97, line 25: Now we can use the MemRef descriptor like this for a temporary solution. I think we should split the MemRef as a separate module (like MemRef in buddy-benchmark) in the future. So can we add a TODO here for future work?

examples/DIPDialect/correlation2D.cpp
```diff
@@ -126,7 +94,8 @@ bool testImplementation(int argc, char *argv[], std::ptrdiff_t x,
   Mat kernel1 = Mat(3, 3, CV_32FC1, laplacianKernelAlign);
 
   // Call the MLIR Corr2D function.
-  _mlir_ciface_corr_2d(input, kernel, output, x, y, 0);
+  dip::Corr2D(input, kernel, output, x, y,
+              dip::BOUNDARY_OPTION::REPLICATE_PADDING);
```
zhanghb97, line 98: Is there an example for `dip::BOUNDARY_OPTION::CONSTANT_PADDING` option? If not, can we add an example for it?

include/dip.hpp
meshtag: What do yo you think about the following?
```
 - include
      - Dialect
      - Target
      - Interface
          - dip
          - dip.h
```

My reason for the change is that I think `buddy` feels a bit more generic inside `Interface` while `dip` specifies our intent of conveying header files for the `DIP` dialect.

meshtag: As discussed, current directory structure is 
```
- include 
    - Dialect
    - Target
    - Interface
        - buddy
            - dip
                - dip.h
                - memref.h
```

Issue #14, opened 2022-02-24
[Build Failed] Edge Detection case doesn't work on macos.


LeiWang1999:
Command:

```bash
/Applications/CLion.app/Contents/bin/cmake/mac/bin/cmake --build /Users/wanglei/buddy-mlir/build --target edge-detection
```

Log:

```bash
[0/1] Re-running CMake...
-- Building with -fPIC
-- 	SSE support - yes
-- 	AVX2 support - yes
-- 	AVX512 support - no
-- Spliting size: 256
/Users/wanglei/buddy-mlir/examples
-- boost include path is : /usr/local/include
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/wanglei/buddy-mlir/build
[2/3] Linking C static library examples/ConvOpt/libConv2D.a
warning: /Library/Developer/CommandLineTools/usr/bin/ranlib: archive library: examples/ConvOpt/libConv2D.a the table of contents is empty (no object file members in the library define global symbols)
[3/3] Linking CXX executable bin/edge-detection
FAILED: bin/edge-detection 
[...]

PR #15, opened 2022-02-24
[Fix] multi system support for the example case edge detect.

LeiWang1999:
Solution for #14


LeiWang1999:
Please CC @zhanghb97

zhanghb97 (approved):
Nice catch! Thanks!

PR #16, opened 2022-02-25
PoolingNhwcSum vectorization

axmat:
- [x] non-strided PoolingNhwcSum
- [x] strided PoolingNhwcSum
- [x] dilated PoolingNhwcSum
- [x] Solve issue with memref and scf
- [x] benchmark and optimize (In another PR for a pooling vectorisation algorithm based on the advanced coefficient broadcasting algorithm)


zhanghb97:
Don't forget to solve the conflict and rebase.
- `mlir/Dialect/StandardOps/IR/Ops.h` -> `mlir/Dialect/Func/IR/FuncOps.h`
- `ReturnOp` -> `func::ReturnOp`

zhanghb97:
I have taken a look at the implementation, the current idea I got from this PR is as the following steps
- allocate memory to hold the element in a "kernel"
- load a vector from the temp memory
- perform the reduction operatoin
- store the result to the output memory

Is my understanding correct?

I think we can load a 2d vector directly and perform the reduction. In this case, we don't need to suffer from the overhead caused by the memory allocation.

axmat:
> I have taken a look at the implementation, the current idea I got from this PR is as the following steps
> 
> * allocate memory to hold the element in a "kernel"
> * load a vector from the temp memory
> * perform the reduction operatoin
> * store the result to the output memory
> 
> Is my understanding correct?
> 
> I think we can load a 2d vector directly and perform the reduction. In this case, we don't need to suffer from the overhead caused by the memory allocation.

Yes, you're right.
Loading a 2d vector directly using memref::SubviewOp gives a "segmentation fault" on large images. The same thing happen when I allocate the memory inside the loop builder. It works fine for small images like memref<1x24x24x1xf32>.

zhanghb97:
> Loading a 2d vector directly using memref::SubviewOp gives a "segmentation fault" on large images. The same thing happen when I allocate the memory inside the loop builder. It works fine for small images like memref<1x24x24x1xf32>.

Have you solved this problem?

Note that you should push the PR after approval for the next time, and use "squash and merge" mode.

We can keep these commits this time, and I will review the PR again. If there is something need to modify, I will let you know.

axmat:
> > Loading a 2d vector directly using memref::SubviewOp gives a "segmentation fault" on large images. The same thing happen when I allocate the memory inside the loop builder. It works fine for small images like memref<1x24x24x1xf32>.
> 
> Have you solved this problem?
> 
> Note that you should push the PR after approval for the next time, and use "squash and merge" mode.
> 
> We can keep these commits this time, and I will review the PR again. If there is something need to modify, I will let you know.

No I couldn’t solve it using memref::SubviewOp because it allocates memory for the pooling window at every iteration inside the loop. So we can’t use memref::SubviewOp here until we figure out how to reuse the allocated memory for the pooling window inside the loop. Allocating and deallocating the pooling window outside the loop builder fix the problem.

zhanghb97:
Actually, you can use the `vector.transfer_read` and `vector.transfer_write` to load/store 2D vector to a memref.

As for the code you have pushed to the main branch, you don't use the 2D strategy, right?

PR #17, opened 2022-02-25
[refactor] better solution for llc triple and attr.

LeiWang1999:
Better solution for #14


LeiWang1999:
Please CC @zhanghb97

zhanghb97 (approved):
Thanks!

PR #25, opened 2022-04-11
create the dialect for digital audio processing

taiqzheng:

PR #31, opened 2022-05-23
Add support for FIR in digital audio processing

SForeKeeper:
This draft pull request is based on taiqzheng's pull request(#25 ). 

To make a series of operations and interfaces that could be put into use, a proper operation design should be considered. 

It seems unnecessary to fully implement all interfaces and functions a library would provide in MLIR. It is proposed that only high-level operations should be implemented and delivered to users as interfaces. 

For example, implementing a fir-filter would need a window function (kaiser, for instance) and the fir function itself. The fir function is simply a 1-D convolution, but implementing a window function would require us to implement the Bessel function, gamma function and other possible high-level functions that are currently not present in MLIR's other dialects. We could, of course, implement those functions in our own DAP dialect, but the current proposal would be only implementing functions that could be lowered to other dialects' existing operations.

To fully utilize the operation we defined and implemented in MLIR, external libraries are needed.


include/Dialect/DAP/DAPOps.td
```diff
@@ -0,0 +1,56 @@
...
+                       Arg<AnyRankedOrUnrankedMemRef, "outputMemref",
+                           [MemRead]>:$memrefO);
+
+  let assemblyFormat = [{
+    $memrefI `,` $memrefK `,` $memrefO attr-dict `:` type($memrefI) `,` type($memrefK) `,` type($memrefO) 
+  }];
+}
+
+#endif // DAP_DAPOPS_TD
+
```
zhanghb97, line 56: Please remove extra spaces on this line.

lib/Conversion/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@ add_subdirectory(ConvVectorization)
 add_subdirectory(LowerBud)
 add_subdirectory(LowerDIP)
 add_subdirectory(LowerRVV)
+add_subdirectory(LowerDAP)
```
zhanghb97, line 5: Please add a new line at the end of the file.

lib/Conversion/LowerDAP/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_mlir_library(LowerDAPPass
+  LowerDAPPass.cpp
+  )
```
zhanghb97, line 3: Please add a new line at the end of the file.

zhanghb97 (commented):
Please add a lowering example to the `examples` directory (from dip to linalg, no need to add the C++ example now).

zhanghb97 (approved):
Thanks!

PR #44, opened 2022-06-09
Search for ANTLR only if samples compilation enabled

ArtemSkrebkov:
## Motivation
Currently, developers do not need ANTLR if they build it with examples disabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,23 +83,26 @@ check_toolchain()
 #-------------------------------------------------------------------------------
 # Antlr Configuration
 #-------------------------------------------------------------------------------
+# NB: currently, ANTLR is used in examples only,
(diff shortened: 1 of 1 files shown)


ArtemSkrebkov:
@zhanghb97 

Please take a look. You had an idea about `BUDDY_FRONTEND_SUPPORT=ON` but I didn't find such the option to re-use.

zhanghb97:
> You had an idea about BUDDY_FRONTEND_SUPPORT=ON but I didn't find such the option to re-use.

Yes! Then let's re-use `BUDDY_DSL_EXAMPLES` as the name.

ArtemSkrebkov:
Right. I missed the fact of using it in DSL samples.
Fixed.

zhanghb97:
Thanks!

PR #46, opened 2022-06-14
ToyDSL

linuxlonelyeagle:
I expand the toydsl use antlr and toy dialect.


examples/ToyDSL/include/toy/Ops.td
```diff
@@ -1,14 +1,18 @@
 //===- Ops.td - Toy dialect operation definitions ----------*- tablegen -*-===//
 //
-// This file is from MLIR toy tutorial Ch-2.
-// TODO: Remove this and directly access the upstream version in the submodule.
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
```
zhanghb97, line 3: Do not copy this LLVM head to here.

zhanghb97 (commented):
Don't remove the makefile, we use the makefile to run the example targets.

PR #23, opened 2022-03-28
Add support for dip.rotate_2d

meshtag:
This PR intends to add support for 2D rotation in DIP dialect via dip.rotate_2d. Following tasks are to be strictly completed before its merger.
WIP Tasks : 
- [x] - Finalise interface via #13 
- [x] - Add [operation description](https://github.com/meshtag/buddy-mlir/blob/bb20823ba1ec4ec670366ff05bbcf9e74ddd2057/include/Dialect/DIP/DIPOps.td#L79) and enhance documentation throughout

PR #49, opened 2022-06-21
Refactor utility functions in separate files

meshtag:
This PR intends to refactor utility functions present inside `lib/Conversion/LowerDIP/LowerDIPPass.cpp` into separate files. These files are created on the basis of their generic utility and DIP specific utility. 

The main purpose of this change is to improve re-usability of these functions across the entire project (developers writing code after the merger of this change should use/improve already available implementation(s) instead of developing similar infrastructure over and over again).

Ref : https://github.com/buddy-compiler/buddy-mlir/pull/23#issuecomment-1155984380


zhanghb97:
@meshtag Please ping me when this PR is ready.

PR #51, opened 2022-06-25
Add C++ wrapper for DAP FIR operation with end-to-end example.

SForeKeeper:
Things done in this patch:
1. A C++ wrapper is purposed and FIR using MLIR, alongside window functions, is implemented.
2. An example of how to use this wrapper is given.
3. Buddy-opt with FIR support is implemented in the previous patch.

Things to do:
1. Modify the AudioFile class, and use its methods to implement an efficient version of AudioContainer. Now it requires multiple copies so the performance won't be good.
2. Implement the Kaiser window and it's math functions.

Things to be discussed:
1. Should we use templates or not. Normally there would be no need to use double(64-bit audios). Even 32-bit floating audio would be an overkill for delivery format. Usually, the audio we found on the web would be quantised to 16 or 24 bits. It should be researched that if processing quantised values would be faster than processing the floating ones. But obviously, the overall algorithm needs to be reimplemented.
2. The overall architecture of this wrapper. Now, this end-to-end fir design is procedural, which means we process data in each function and pass processed data. It would possibly have a negative impact on higher-level optimizations. It would allow us to dig deeper if we make it functional, which means we specify a graph and then propagate it with actual data, instead of processing data step by step.
3. Paths in test files. Would it be better if use the path from the repository root instead of using `../../` .
4. MemRef as base class. We shouldn't use MemRef as base class, the destructor needs to be virtual so inherited classes could call destructors properly, but then the memory layout would change and mlir c interface won't work correctly.


include/Interface/buddy/dap/dsp/fir.h
```diff
@@ -0,0 +1,74 @@
...
+template <typename T, size_t N>
+void fir(Audio<float, N> *input, MemRef<T, N> *filter,
+         Audio<float, N> *output) {
+  if (N != 1)
+    assert(0 && "Only mono audio is supported for now.");
+  detail::_mlir_ciface_conv1d_buddy(input, filter, output);
+}
+} // namespace dap
+
+#endif // INCLUDE_DAP_DSP_FIR
```
zhanghb97, line 74: Add an empty line at the end of the file.

include/Interface/buddy/dap/dsp/math.h
```diff
@@ -0,0 +1,62 @@
...
+// Basic math functions
+template <typename T> T sinc(T x) { return sin(x) / x; }
+
+template <typename T> T besseli0(T x) {
+  assert(0 && "Not implemented.");
+  // return besseli(0,x);
+  return 0;
+}
+
+//  float besseli(float _nu,
```
zhanghb97, line 37: If the function is unnecessary, please remove it.

include/Interface/buddy/dap/dsp/window.h
```diff
@@ -0,0 +1,158 @@
+//===- window.h
```
zhanghb97, line 1: Wrong format.

include/Interface/buddy/dap/dsp/window.h
```diff
@@ -0,0 +1,158 @@
...
+  case WINDOW_TYPE::TRIANGULAR:
+    if (!args)
+      assert(0 && "Argument not provided.");
+    return _window_triangular<T>(i, len, args[0]);
+  }
+}
+} // namespace detail
+} // namespace dap
+
+#endif // INCLUDE_DAP_DSP_WINDOW
```
zhanghb97, line 158: Add an empty line at the end of the file.

lib/Interface/core/CMakeLists.txt
```diff
@@ -5,3 +5,5 @@ if(BUDDY_ENABLE_OPENCV)
   include_directories(${OpenCV_INCLUDE_DIRS})
   add_library(ImageContainer ImageContainer.cpp)
 endif()
+
+add_library(AudioContainer AudioContainer.cpp)
```
zhanghb97, line 9: Add an empty line at the end of the file.

tests/Interface/core/CMakeLists.txt
```diff
@@ -13,3 +13,9 @@ _add_test_executable(buddy-image-container-test
     ${OpenCV_LIBS}
 )
 endif()
+
+_add_test_executable(buddy-audio-container-test
+  AudioContainerTest.cpp
+  LINK_LIBS
+    Container AudioContainer
+)
```
zhanghb97, line 21: Add an empty line at the end of the file.

lib/Interface/core/Container.cpp
```diff
@@ -193,6 +193,8 @@ template <typename T, std::size_t N> T &MemRef<T, N>::operator[](size_t index) {
 template <typename T, std::size_t N> void MemRef<T, N>::setStrides() {
   assert((N > 0) && "Invalid container number of dims");
   strides[N - 1] = 1;
+  if (N<2)
```
zhanghb97, line 196: `N<2` -> `N < 2`

SForeKeeper:
The math functions and ugly annotations of certain functions remain untouched, will implement them next.

zhanghb97 (commented):
- When I run the `./FirLowpass`, it reports the segmentation fault.
- Please add a section (how to build & how to run) to `buddy-mlir/examples/README.md`.

zhanghb97 (approved):
Thanks!

Issue #19, opened 2022-03-07
Add caching mechanism for llvm build in GitHub Actions script


meshtag:
**Is your feature request related to a problem? Please describe.**
GitHub actions builds are terribly slow due to the initiation of fresh llvm build process on each commit.

**Describe the solution you'd like**
We should cache the llvm build data daily (since we update llvm dependency daily). This would enable faster CI builds when multiple commits are pushed during the same day.

**Additional context**
Ref : https://github.com/buddy-compiler/buddy-mlir/blob/297956e034cc55262b2e97033e9a0d4fafbf495b/.github/workflows/TestBuild.yml#L12

Since this issue doesn't require any significant knowledge of codebase (GitHub actions knowledge will be required), I am marking it as a Good first issue.

RishiKumarRay:
@meshtag will this work 
  - uses: actions/cache@v2
    with:
      path: | 
        path/to/dependencies
        some/other/dependencies 
      key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}

meshtag:
Thanks for your interest @RishiKumarRay !

> will this work

You can check it by adding a branch from your fork [here](https://github.com/buddy-compiler/buddy-mlir/blob/297956e034cc55262b2e97033e9a0d4fafbf495b/.github/workflows/TestBuild.yml#L5) .

Feel free to create a PR once your solution works :)

RishiKumarRay:
@meshtag sure i will check this and raise pr

ArtemSkrebkov:
Probably, can be closed after merging a PR #57

zhanghb97:
Thanks @ArtemSkrebkov

PR #54, opened 2022-06-30
Update scripts to run RISC-V emulator using LIT

ArtemSkrebkov:
# Motivation
Many of the examples from `MLIRVector`, `RVVExperiment` can be used as regression tests. It will help to test them regularly and will save the cost of maintenance.

# Opens
* Not ready to be integrated in CI since RISC-V environment is time-consuming to set up. Would need to have https://github.com/buddy-compiler/buddy-mlir/issues/19 resolved and use the same approach for the environment.
* Currently, if there is no RISC-V environment, the corresponding tests will fail. We could make those tests optional and do not run them if the environment is not available.

### DIFF
--- a/.github/workflows/TestBuild.yml
+++ b/.github/workflows/TestBuild.yml
@@ -27,5 +27,7 @@ jobs:
         with:
           path: llvm
           key: build-${{ steps.llvm-build-dir.outputs.commit }}
+      - name: Build RVV enviroment
+        run: bash ./thirdparty/build-rvv-env.sh
       - name: Test build
         run: bash ./tests/Actions/TestBuild.sh  llvm/build-${{ steps.llvm-build-dir.outputs.commit }}
--- a/tests/lit.cfg.py
+++ b/tests/lit.cfg.py
@@ -45,23 +45,28 @@ config.excludes = ['Inputs', 'Examples', 'CMakeLists.txt', 'README.txt', 'LICENS
 # test_source_root: The root path where tests are located.
 config.test_source_root = os.path.dirname(__file__)
 
-# test_exec_root: The root path where tests should be run.
-config.test_exec_root = os.path.join(config.buddy_obj_root, 'tests')
+# tools_dir: tool directories to be used when testing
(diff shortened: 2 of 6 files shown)


tests/lit.cfg.py
```diff
@@ -45,19 +45,28 @@
 # test_source_root: The root path where tests are located.
 config.test_source_root = os.path.dirname(__file__)
 
-# test_exec_root: The root path where tests should be run.
-config.test_exec_root = os.path.join(config.buddy_obj_root, 'tests')
+# tools_dir: tool directories to be used when testing
 config.buddy_tools_dir = os.path.join(config.buddy_obj_root, 'bin')
+config.qemu_tools_dir = "/home/artemy/workspace/repos/buddy-mlir/thirdparty/qemu/build/riscv64-linux-user"
```
ArtemSkrebkov, line 50: Remove absolute path

.github/workflows/TestBuild.yml
```diff
@@ -27,5 +27,7 @@ jobs:
         with:
           path: llvm
           key: build-${{ steps.llvm-build-dir.outputs.commit }}
+      - name: Build RVV enviroment
+        run: bash ./thirdparty/build-rvv-env.sh
```
ArtemSkrebkov, line 31: Looks like this script does not satisfy resources restrictions. Logs:
https://pipelines.actions.githubusercontent.com/serviceHosts/a2afae62-3cc1-4bba-8025-9c9e34c59e8a/_apis/pipelines/1/runs/286/signedlogcontent/2?urlExpires=2022-08-03T14%3A35%3A18.3447346Z&urlSigningMethod=HMACV1&urlSignature=wrCuK8O%2BbleXbepZnJqXJKvNXiS3HkcYJwARdsL0Dus%3D

meshtag:
What is the status of this PR? @ArtemSkrebkov

ArtemSkrebkov:
Thanks for reaching out.

Sorry, I was busy with other things. If you think that it still can be useful, I can finalize it in spare time.

PR #55, opened 2022-07-10
Add support for resizing in DIP dialect

meshtag:
This PR intends to add support for resizing images with `DIP` dialect. As of now, support for two interpolation schemes is developed, namely `nearest neighbour interpolation` and `bilinear interpolation`. 

These specific types of interpolation schemes are encapsulated in modular functions, this strategy is used for the purpose of keeping the implementation as extensible as possible which should make it easier for developers to add support for novel algorithms to handle existing operations in `DIP` dialect in future.


zhanghb97:
Merged! Thanks!
The next step is to use the unified MemRef container.

PR #59, opened 2022-07-15
An example for dialect interface: BudInlinerInterface

taiqzheng:
**Overview:**

An example for dialect interface.

**Detail:**

- Inline interface is a kind of dialect interfaces, which may be used in transformations & analyses(mlir-opt --inline).

- This PR intended to show how to create an inline interface(and what ever kinds of dialect interfaces), as well as how to use it(buddy-opt --inline). 

- Path: buddy-mlir/examples/BudDialect/TestInlineInterface.mlir.

- Use "make bud-inline-interface-run" to try it out.


zhanghb97 (commented):
Please rename the makefile target:

`bud-inline-interface-run` ->` bud-inline-interface-lower`

taiqzheng:
The makefile target have changed to `bud-inline-interface-lower`.
The conflict have solved.

zhanghb97 (approved):
Thanks!

PR #62, opened 2022-07-17
Use modified Memref containers in DIP dialect examples

meshtag:
This PR intends to add support for the novel Memref container implementation in all DIP dialect examples instead of the previous one.


zhanghb97 (approved):
Thanks!

PR #75, opened 2022-09-20
[OSPP]frontendgen

linuxlonelyeagle:
Add a frontend named frontendgen and some examples to use it.you can use the examples in the 
**buddy-mlir/examples/FrontEndGen**.


examples/FrontEndGen/makefile
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+BUDDY_FRONTEND_GEN := ../../build/bin/buddy-frontendgen
+
+frontendgen-emit-ast:
+	@${BUDDY_FRONTEND_GEN} -f ./example.fegen -emit=ast
+
+frontendgen-emit-antlr:
+	@${BUDDY_FRONTEND_GEN} -f ./example.fegen -emit=antlr -g Toy
+
+frontendgen-emit-visiror:
```
zhanghb97, line 10: typo: frontendgen-emit-visiror -> frontendgen-emit-visitor

zhanghb97 (changes requested):
- Don't try to center filenames in the header. The filename in the header is like this:
```
//===- xxx.h ---------------------------------------------------===//
```

- Please eliminate the warnings
    -  "xxx" will be initialized after
    -  variable 'xxx' set but not used

zhanghb97 (approved):
Thanks!

PR #81, opened 2022-10-15
[thirdparty] disable LLVM zstd support rvv build scripts

inclyc:
LLVM D128465 introduced zstd as an optional compression method in LLVM. However with in the cross-compile environment there is no such library by default. This patch adds a cmake option that disable zstd support in build-rvv-env.sh.

Link: https://reviews.llvm.org/D128465


zhanghb97 (approved):
Thanks!

PR #83, opened 2022-11-07
Improve CBConvVectorization Lowering pass

meshtag:
This PR intends to improve `CBConvVectorization` lowering pass by adding support for neglecting null kernel elements. This should save us some extra assembly instructions (for FMA calculation, tail processing, etc.) in cases where we encounter a null kernel element. 

I am attaching benchmark results with and without this modification to illustrate the above behaviour. 

Before modification : 649 iterations
Screenshot from 2022-11-07 11-58-32

After modification : 1509 iterations
Screenshot from 2022-11-07 23-15-15


zhanghb97:
Nice improvement!

PR #84, opened 2022-11-09
[DIP] replace C-style variadic function with std::vector

Origami404:
The `checkDIPCommonTypes` function uses C-style variadic arguments only for passing variadic numbers of the same type arguments, which should be avoided in C++. It is recommended to replace this usage with `std::initializer_list` or `std::vector`.

This PR also fixes the compilation failure in `clang++`, which will give a warning (as an error) for non-POD type `Value` in `va_arg`.

### DIFF
--- a/include/Utils/DIPUtils.h
+++ b/include/Utils/DIPUtils.h
@@ -22,6 +22,7 @@
 #ifndef INCLUDE_UTILS_DIPUTILS_H
 #define INCLUDE_UTILS_DIPUTILS_H
 
+#include "DIP/DIPOps.h"
 #include "Utils/Utils.h"
 #include <stdarg.h>
 
@@ -197,7 +198,7 @@ void traverseImagewBoundaryExtrapolation(
 
 // Function for applying type check mechanisms for all DIP dialect operations.
 template <typename DIPOP>
-DIP_ERROR checkDIPCommonTypes(DIPOP op, size_t numArgs, ...);
+DIP_ERROR checkDIPCommonTypes(DIPOP op, const std::vector<Value> &args);
 } // namespace dip
 } // namespace buddy
 
--- a/include/Utils/DIPUtils.h
+++ b/include/Utils/DIPUtils.h
@@ -22,7 +22,6 @@
 #ifndef INCLUDE_UTILS_DIPUTILS_H
 #define INCLUDE_UTILS_DIPUTILS_H
 
-#include "DIP/DIPOps.h"
 #include "Utils/Utils.h"
 #include <stdarg.h>
 
--- a/lib/Conversion/LowerDIP/LowerDIPPass.cpp
+++ b/lib/Conversion/LowerDIP/LowerDIPPass.cpp
@@ -72,7 +72,7 @@ public:
 
     auto inElemTy = input.getType().cast<MemRefType>().getElementType();
     dip::DIP_ERROR error = dip::checkDIPCommonTypes<dip::Corr2DOp>(
-        op, 4, input, kernel, output, constantValue);
+        op, {input, kernel, output, constantValue});
 
     if (error == dip::DIP_ERROR::INCONSISTENT_TYPES) {
       return op->emitOpError() << "input, kernel, output and constant must "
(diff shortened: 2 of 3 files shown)


meshtag:
Can you please specify the `clang` version which leads to the mentioned failure

meshtag:
I was able to reproduce the error. 
(Ref : https://github.com/meshtag/buddy-mlir/actions/runs/3439561667/jobs/5737001087#step:6:4405 )

Origami404:
> Can you please specify the `clang` version which leads to the mentioned failure

Sorry for the late reply, I use clang 14 on my machine. Here is the output of `clang -v`:

```
clang version 14.0.5 (Fedora 14.0.5-1.fc36)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/12
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-redhat-linux/12
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
```

meshtag:
These modifications seem to work on `macos-latest` configuration.
(Ref : https://github.com/meshtag/buddy-mlir/actions/runs/3442213404/jobs/5742558798 )

include/Utils/DIPUtils.h
```diff
@@ -22,6 +22,7 @@
 #ifndef INCLUDE_UTILS_DIPUTILS_H
 #define INCLUDE_UTILS_DIPUTILS_H
 
+#include "DIP/DIPOps.h"
```
meshtag, line 25: Why are you including this file?

meshtag (changes requested):

include/Utils/DIPUtils.h
Origami404: Sorry, my `clangd` automatically includes this file but I haven't noticed it. It is unnecessary for compilation, but removing it will cause my `clangd` to produce a ton of errors for this file.

The real problem is that the `Utils/Utils.h` file is not self-contained. `Utils/Utils.h` is which needs this inclusion. I will remove this #include and add one to `Utils/Utils.h` in another PR.

meshtag: Can you please share the error log you get when you don't include that file.

Origami404: Screenshot from 2022-11-11 23-19-13

meshtag: I think this is an issue related only to intellisense, as [this](https://github.com/meshtag/buddy-mlir/actions/runs/3442213404/jobs/5742558798) run compiled well.  
(Ref : https://stackoverflow.com/questions/31943634/visual-studio-2015-or-2017-shows-intellisense-errors-but-solution-compiles )

Origami404: > Sorry, my `clangd` automatically includes this file but I haven't noticed it. It is unnecessary for compilation, but removing it will cause my `clangd` to produce a ton of errors for this file.

Yes, you are right. It's **only** an IntelliSense error in **`clangd`**, not about using **`clang`** to compile the whole project. I can also successfully compile the project without that `#include` locally as well.

> The real problem is that the `Utils/Utils.h` file is not self-contained. `Utils/Utils.h` is which needs this inclusion. I will remove this #include and add one to `Utils/Utils.h` in another PR.

But the IntelliSense error indicates a possible problem in `Utils/Utils.h`. A header file that is not self-contained will depend on the order of inclusion in other files. For example:

```cpp
// in file A.cpp that needs to include Utils.h
// this will work
#include "DIP/DIPOps.h"
#include "Utils/Utils.h"

// in file B.cpp that needs to include Utils.h
// this will NOT work
#include "Utils/Utils.h"
#include "DIP/DIPOps.h"
```

As we include the `Utils/Utils.h` at the beginning of `Utils/DIPUtils.h`, we also "pollute" the `Utils/DIPUtils.h`, making it not self-contained as well. In fact, the whole project can compile is only because we have not written any C++ file (non-header file) that will include `Utils/DIPUtils.h` as the first file it will include.

In my opinion, making header files self-contained should be a best practice for C/C++ projects. So I said I may open another PR to fix this (or maybe more non self-contained header files in the project). You may also refer [this SO answer](https://stackoverflow.com/a/1892054).

Anyway, this problem is **not** related to the goal of the current PR (fixing the compilation failure under `clang`), we may discuss it later and I also push a new commit to delete the unnecessary inclusion. I am sorry if my previous words were not clear enough.

meshtag: Alright

meshtag (approved):
Thanks!

PR #89, opened 2022-12-02
Improve codes in CGModule

Origami404:
# Improve codes in CGModule

This PR improves codes in `frontend/FrontendGen/lib/CGModule.cpp`.

## Fixed typos in this file:

A list of typos is below:

- `grammarNmae` -> `grammarName`
- `formuls` -> `formulas`
- `gramarName` -> `grammarName`
- `indexs` -> `indices`
- `opArgments` -> `opArguments`
- `resultsmap` -> `resultsMap`
- `argmentsMap` -> `argumentsMap`
- `buildrOp` -> `builderOp`
- `getbulderIdxs` -> `getBuilderIndices`

## Fixed wrong index type for storing the result of `string::find`

In lines `19` and `47`, `start` and `end` are used to store the result of `std::string::find`, which returns a `std::size_t`. The `std::string::npos` can never be stored in an `uint`, which causes the conditions (e.g. `start == std::string::npos`) below never satisfy.

## Fixed passing temporary string objects to StringRefs

In the original code, we have:

```cpp
llvm::SmallVector<llvm::StringRef> opArgments;

//...

os << "arg" << index << ";\n";
llvm::StringRef arg("arg" + std::to_string(index));
opArgments.push_back(arg);
```

Here comes a problem: expression `"arg" + std::to_string(index)` will create a temporary string object, which may be destroyed at the end of the expression. But, the StringRef in `opArgments` still keeps a pointer into this string's data, which will become a dangling pointer.

This PR introduced a new vector named `tmpStrings` to keep all the temporary strings, ensuring their lifetime is at least as long as `opArgments`.

An alternative approach is to change `opArgments` to a `SmallVector<std::string>`. But this may cause a performance problem because most of the elements in `opArgments` seem just a reference to other strings.

Because I am not familiar with this part of the code, there may be some other lifetime problems. A deeper review may be needed.

### DIFF
--- a/frontend/FrontendGen/include/AST.h
+++ b/frontend/FrontendGen/include/AST.h
@@ -60,7 +60,7 @@ public:
   llvm::SmallVector<llvm::StringRef, 4> getBuilderNames() {
     return this->builderNames;
   }
-  llvm::SmallVector<int> getbulderIdxs() { return this->builderIdxs; }
+  llvm::SmallVector<int> getBuilderIndices() { return this->builderIdxs; }
 };
 
 /// This class is used to mark the node in the generator as a rule, and can also
--- a/frontend/FrontendGen/include/CGModule.h
+++ b/frontend/FrontendGen/include/CGModule.h
@@ -22,17 +22,17 @@
 namespace frontendgen {
 
 /// TypeMap is used to store type maps.The cppMap is used to map c++ types,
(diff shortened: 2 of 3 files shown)


zhanghb97:
@Origami404 Thanks!
This is the initial demo of the Frontend Generator idea, which is a part of the OSPP project.
We need to discuss the details of this direction and write more tests.

PR #91, opened 2022-12-06
add `-fsanitize=address` in CMakeLists.txt

Pzzzzz5142:
I cannot compile this project on neither macOS nor ArchLinux until I explicitly add `-fsanitize=address` in CMakeLists.txt. Although this may due to my local compile env setup, I find that llvm-project has added it. So I think this setup may be included in CMakeLists.txt rather than in local env setup.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,9 @@ project(buddy-mlir LANGUAGES CXX C)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
 
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+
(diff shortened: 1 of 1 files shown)


SForeKeeper:
Hi, I have a few questions on your problem and the pull request itself.

1. Could you please provide more detailed information about why the project can't be built on your device? I have succeeded in compiling it on macOS 12.6 with Apple Silicon, and Ubuntu Linux on both arm64 and amd64 platforms.
2. Since ASAN typically slows a program down by 2x, it seems a bit irrational adding `-fsanitize=address` to the whole project. According to [Building LLVM with CMake](https://llvm.org/docs/CMake.html), `LLVM_USE_SANITIZER` has a default value of an empty string, thus llvm-project is not configured with any sanitizers turned on. 

I'd suggest you check your libraries, if buddy-mlir is configured to use some instrumented libraries, it could be necessary to turn sanitizers on to compile.

Pzzzzz5142:
Sorry, it turns out that the problem was caused by using my local version of mlir. Now everything compiles normally.

PR #93, opened 2022-12-15
[buddy-mlir build] llvm::Optional => std::optional

Origami404:
LLVM project is incremental changing and the deprecate `llvm::Optional` now, using `std::optional` instead. It results in many, many deprecation warnings when compiling buddy-compiler. This PR aims to do a whole project upgrade from `llvm::Optional` to `std::optional`.

Change list:

- `llvm::None` => `std::nullopt`
- `llvm::Optional` => `std::optional` 

Sorry that I can not think out a `[part]` for the commit message because I could not find a proper one from git history. I will be happy to fix it if anyone could suggest a prefix.


meshtag (changes requested):
Can you please attach the llvm release notes which dictate that `llvm::Optional` is now deprecated. I was able to observe deprecation messages for `llvm::None` by simply using `ninja check-buddy | grep llvm::None` on a fresh build, but couldn't find deprecation warnings for `llvm::Optional` in the same manner. 

PS: Perhaps you can use the tag `[buddy-mlir build]` as prefix for the commit message, since this PR tries to improve the overall build of the project. (/cc @zhanghb97)

Origami404:
> Can you please attach the llvm release notes which dictate that `llvm::Optional` is now deprecated. 

Sorry for my misleading description. I mean to say that:

1. Incremental migrating from `llvm::Optional` to `std::optional` is processing.
2. Being a part of the migrating process, `llvm::None` is deprecated
3. `llvm::Optional` is not deprecated yet

For 1, the [LLVM weekly #465](https://discourse.llvm.org/t/llvm-weekly-465-november-28th-2022/66860) notes that:

> Many, many patches were landed to convert uses of LLVM’s Optional class to std::optional. e.g. [edace86](https://reviews.llvm.org/rGedace862f63a), [b54cd3f](https://reviews.llvm.org/rGb54cd3f09c96).

And by `git log --grep='Use std::optional'` in the LLVM repo, you can find many commits doing this. IMO, this kind of migrating should be done now when we don't have so much usage of `llvm::Optional` in buddy-compiler yet.

For 2, the deprecation warnings have indicated that.

For 3, I checked `llvm/ADT/Optional.h`, recent commits, and the discourse.

meshtag:
Please use `#include <optional>` at relevant places.

meshtag (approved):

Origami404:
Oh I am so sorry to say that I have completely forgot this PR after I recovered from nCov-2019. Thanks @meshtag to help me fix this!

PR #94, opened 2022-12-16
Add lib

linuxlonelyeagle:
I added some libs that llc needs and fixed the buddy-llc.buddy-llc can run successfully!


zhanghb97 (approved):
Thanks!

PR #96, opened 2022-12-17
[Container] Fix implicit conversion between 'long unsigned' and 'int'

Panhaolin2001:
This PR intends to suppress the overflow warning by preventing an implicit conversion from "long unsigned int" to "int".

### DIFF
--- a/lib/Interface/core/Container.cpp
+++ b/lib/Interface/core/Container.cpp
@@ -211,8 +211,9 @@ template <typename T, std::size_t N> void MemRef<T, N>::setStrides() {
   strides[N - 1] = 1;
   if (N < 2)
     return;
-  for (int i = N - 2; i >= 0; i--) {
-    strides[i] = strides[i + 1] * sizes[i + 1];
+  // Prevent implicit conversions between unsigned and signed
+  for (std::size_t i = N - 1; i > 0; i--) {
+    strides[i - 1] = strides[i] * sizes[i];
   }
 }
 


lib/Interface/core/Container.cpp
```diff
@@ -211,7 +211,7 @@ template <typename T, std::size_t N> void MemRef<T, N>::setStrides() {
   strides[N - 1] = 1;
   if (N < 2)
     return;
-  for (int i = N - 2; i >= 0; i--) {
+  for (int i = static_cast<int>(N - 2); i >= 0; i--) {
```
zhanghb97, line 214: What warning will be reported now?
I think we can change `int i = ...` to `size_t i = ...`.

Panhaolin2001: if using origin code 'int i = N - 2 ', the error is "warning: overflow in conversion from ‘long unsigned int’ to ‘int’ changes value from ‘18446744073709551615’ to ‘-1’ [-Woverflow]".
image
if replace 'int i = N - 2' with ''size_t i = N - 2'',it can build successfully. But, the executable file "vectorization-benchmark" will cause 'segmentation fault'.
image

meshtag: This seems weird, I'll try to reproduce the segmentation fault error.

meshtag: @Panhaolin2001 , can you share your local `buddy-mlir` and `buddy-benchmark`'s latest git commit id, which results in the above mentioned `segmentation fault`.

Panhaolin2001: @meshtag The case that caused this error is a PR in buddy- benchmark ：_https://github.com/buddy-compiler/buddy-benchmark/pull/50_
BTW, the MemRef I used whose dimension is one.

Panhaolin2001: The error occurred in two other benchmarks in the Vectoration folder. Both had a MemRef dimension of 2. When this code was executed:
```
for (std::size_t i = N - 2; i >= 0; i--) {
    strides[i] = strides[i + 1] * sizes[i + 1];
  }
```
The initial value of i was 0, and after a round of loops i became -1. After converting i to an unsigned integer (i = 18446744073709551615), the loop condition was met and the loop body was executed again.
I think there are two solutions to this issue. The first is to cast the size_t type of N-1 to int. The second is to change the loop index:
```
for (std::size_t i = N - 1; i > 0; i--) {
    strides[i - 1] = strides[i] * sizes[i];
  }

```

meshtag: `std::size_t` actually corresponds to unsigned integer type. 
(Ref : https://en.cppreference.com/w/cpp/types/size_t#:~:text=std%3A%3Asize_t%20is%20the,(since%20C%2B%2B11).)
The last iteration of this loop `for (std::size_t i = N - 2; i >= 0; --i)` might actually lead to the assignment of a negative value (-1) to the variable `i`. This will definitely cause issues. 

@Panhaolin2001, can you check if this is the case by making some changes to verify this with your PR on your local system.

meshtag: Changing the loop index (with a comment on that location explaining the reason) makes more sense to me.

Panhaolin2001: OK,thanks, i will change the loop index now.

meshtag (approved):
LGTM 
(/cc @zhanghb97)

PR #100, opened 2022-12-28
[RVV] Add Rsqrt operation

xlinsist:
## Summary
1. Add Rsqrt operation for RVV Dialect.
2. Add Resqrt lowering example from Math Dialect to X86vector Dialect for testing and reference.

## Motivation
When expressing deep learning models in MLIR ecosystem, **Rsqrt operation** (Reciprocal of the square-root) frequently shows up to participate in the calculations of BN or LN(Batch or Layer Normalization) in the form of `tosa.rsqrt` or `math.rsqrt`. In normal lowering pipeline, they will be divided into two operations, `llvm.intr.sqrt` and `llvm.fdiv`. It is neither efficient nor elegant.  

To address the problem, x86vector dialect supports [x86vector.avx.rsqrt](https://mlir.llvm.org/docs/Dialects/X86Vector/#x86vectoravxrsqrt-mlirx86vectorrsqrtop) to directly target [specific AVX instruction](https://uops.info/html-instr/VRSQRTPS_XMM_XMM.html), but similar solution hasn't be found in RISC-V so far. RVV has instruction **`vfrsqrt7`** to specifically deal with the square-root reciprocal, so I think adding the related operation in RVV dialect can facilitate supports for deep learning and other mathematical models in RISC-V.

## Work
1. An unmasked version of rsqrt as well as its intrinsic operation is inplemented, naming 'RsqrtOp' and 'IntrFrsqrt7Op' in _RVV.td_ respectively('RVV' prefix is omited for simplicity). As for the masked version, I am digging out suitable potential usages.
2. Go back to `math.rsqrt`, it's already supported to be lowered to `x86vector.avx.rsqrt` with `test-math-polynomial-approximation` pass, so I add the related example here for testing and reference.

## Rsqrt Example in RVV Dialect

Define in RVV dialect: 
`%res_rvv = rvv.rsqrt %load_vec_f32, %vl8 : vector<[8]xf32>, index`

Lower to intrinsic operation:
`%31 = "rvv.intr.vfrsqrt7"(%30, %28, %23) : (vector<[8]xf32>, vector<[8]xf32>, i64) -> vector<[8]xf32>`

Translate to llvm IR:
`%7 = call <vscale x 8 x float> @llvm.sqrt.nxv8f32(<vscale x 8 x float> %2)`

Comile and run. A demo result is shown below while the first line represents the output of `rvv.rsqrt` and the second line represents that of `math.rsqrt` given the same input(with acceptable error).
```
[inf,  0.996094,  0.703125,  0.574219,  0.498047,  0.445312,  0.40625,  0.376953,  0,  0,  
inf,  1,  0.707107,  0.57735,  0.5,  0.447214,  0.408248,  0.377964,  0,  0]
```

### DIFF
--- a/examples/MLIRMath/makefile
+++ b/examples/MLIRMath/makefile
@@ -36,3 +36,39 @@ math-atan2-run:
 		-reconcile-unrealized-casts | \
 	${MLIR_CPU_RUNNER} ${OPT_FLAG} -e main -entry-point-result=void \
 		-shared-libs=${MLIR_RUNNER_UTILS} -shared-libs=${MLIR_C_RUNNER_UTILS}
+
+math-rsqrt-lower:
+	@${MLIR_OPT} ./math-rsqrt.mlir \
+		-convert-math-to-llvm -convert-func-to-llvm \
+		-o ./log.mlir
+
+math-rsqrt-translate:
+	@${MLIR_OPT} ./math-rsqrt.mlir \
+		-convert-math-to-llvm -convert-func-to-llvm | \
+	${MLIR_TRANSLATE} \
+		-mlir-to-llvmir -o log.ll
+
+math-rsqrt-lower-x86:
+	@${MLIR_OPT} ./math-rsqrt.mlir \
+		-test-math-polynomial-approximation=enable-avx2 \
+		-convert-vector-to-llvm="enable-x86vector" \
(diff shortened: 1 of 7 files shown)


zhanghb97 (approved):
LGTM, thanks!

PR #103, opened 2023-01-13
[Backend]Add gemmini instructions

linuxlonelyeagle:
I have added some instructions to support gemmini, and updated the RISCVBuddyExt example. You can use the makefile to test the examples.This example transfers the value of array A to array B, and outputs the value of array B.


linuxlonelyeagle:
If you want to try this example.You have to build [gemmini](https://github.com/ucb-bar/gemmini).When you have successfully built gemmini, you should enter the `chipyard` directory, then run the following command to set environment variables.
```
source env.sh 
```
Here are the makefile options.
* rv-buddy-mvin-mvout-asm
 Generate asm file.
* rv-buddy-mvin-mvout-obj
 Generate obj file.
* rv-buddy-mvin-mvout-exe
 Generate executable file.
* rv-buddy-mvin-mvout-run 
 Run the spike emulator to run the example.

Now you can test this example.You can run the following command.
```
make rv-buddy-mvin-mvout-obj
make rv-buddy-mvin-mvout-exe
make rv-buddy-mvin-mvout-run
```

linuxlonelyeagle:
I added the rv-buddy-matrix-add example.In this example, the values ​​in array A and array B are added by software and stored in the gold array. Then transfer the arrays A and B into gemmini, then use gemmini to add the matrixs, and finally output the result, and compare the final result with the gold array.At the same time I also provide makefile options,similar to the previous example, you can test this example.

linuxlonelyeagle:
I added the rv-buddy-transpose example and preload and computePreloaded `Intrinsic` in llc .In this example.The `systolic array` in gemmini is used.First load the array C and array D as 0 matrix,then the A array(In array) is loaded and transposed by hardware, then multiply it with the B matrix(identity matrix).You can see the [formulas](https://github.com/ucb-bar/gemmini#core-matmul-sequences).

linuxlonelyeagle:
I added the `rv-buddy-mvin-padded` example,this example transfers a 13x14 matrix to gemmini and outputs it to the output matrix(16x16 matrix),this example uses gemmini's pading feature.You can test it using makefile.

linuxlonelyeagle:
I added the rv-buddy-mvout-padded example,this example transfers a 16x16 matrix to gemmini and outputs it to the output matrix(13x13 matrix),this example uses gemmini's pading feature.You can test it using makefile.

linuxlonelyeagle:
I added the `rv-buddy-matmuls-padded` example.Array A is 14 * 16 and array B is 16 * 15. arrays C and D are both 14 * 15. This example multiplies the A and B arrays and then adds them to the C array, then transfers them to the D array, and then the D array is compared to the gold array, which is calculated by software simulation. This example uses both `OUTPUT-STATIONARY` and `WEIGHT-STATIONARY` for the calculation.You can find some [information](https://github.com/ucb-bar/gemmini#systolic-array-and-transposer).

linuxlonelyeagle:
I  added the `rv-buddy-tile-matmul-ws` example and `loop instruction`.This example has four arrays A,B,C,D and they are all 64*64. Then the result of A*B+D is stored in C.

linuxlonelyeagle:
The test cases for the `compute_ccumulated instruction` will be added later.

linuxlonelyeagle:
I added `rv-buddy-conv` example and `loop conv instruction`.This is an example of convolution.output is calculated with cpu, and output_mat is calculated with gemmini.

zhanghb97 (approved):
Thanks!

PR #112, opened 2023-02-03
[AudioContainer] Clipped the output value in range -1 to 1

amanchhaparia:
For floating format only values between -1 and 1 are legal. The third party encoder used for Audio container does not handle the values beyond the range and produces undesired results. This PR intends to resolve this issue.


SForeKeeper (approved):
Only clamping float values for now.

meshtag:
@amanchhaparia, please resolve the merge conflict and update the PR.

PR #113, opened 2023-02-05
Examples of how to use mlir dialects to implement gemm.

hxdtest:


meshtag:
Please have a look at [this](https://buddy-compiler.github.io/Pages/ContributorGuide.html#pull-requests) and make appropriate changes.

zhanghb97:
@hxdtest Thanks for providing the examples!
(1) Please reformat your code as @meshtag said above.
(2) How do you measure the execution time? Can we add a GFLOPS report to the example? (see [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/benchmark/conv2d-buddy-template#L69) for an example)
(3) Have you considered adding a vectorization step to the example?

examples/MLIRGEMM/gemm.mlir
```diff
@@ -0,0 +1,289 @@
...
+
+    // naive muatrix multiplication using linalg dialect 
+    func.func @matmul(%a : memref<?x?xf64>, %b : memref<?x?xf64>, %c : memref<?x?xf64>) {
+      linalg.matmul 
+        ins(%a, %b: memref<?x?xf64>, memref<?x?xf64>)
+       outs(%c:memref<?x?xf64>)
+      return
+    }
+
+    // naive muatrix multiplication using affine dialect
```
Origami404, line 25: maybe change the comment to "naive matrix multiplication **with fix length** using affine dialect" to distinct `matmul1` from `matmul2`?

examples/MLIRGEMM/gemm.mlir
```diff
@@ -0,0 +1,289 @@
...
+#map14 = affine_map<(d0) -> (d0 * 8)> 
+#map15 = affine_map<(d0) -> (d0 * 8 + 8)> 
+#map16 = affine_map<(d0) -> (d0 * 16)> 
+#map17 = affine_map<(d0) -> (d0 * 16 + 16)> 
+
+module{
+
+    func.func private @printMemrefF64(memref<*xf64>)
+
+    // naive muatrix multiplication using linalg dialect
```
Origami404, line 17: typo `muatrix` -> `matrix`

examples/MLIRGEMM/gemm.mlir
```diff
@@ -0,0 +1,289 @@
...
+        ins(%cf2 : f64)
+       outs(%C2:memref<?x?xf64>)
+
+       %print_A = memref.cast %A : memref<?x?xf64> to memref<*xf64>
+       //call @printMemreff64(%print_A) : (memref<*xf64>) -> ()
+
+       %print_B = memref.cast %B : memref<?x?xf64> to memref<*xf64>
+       //call @printMemreff64(%print_B) : (memref<*xf64>) -> ()
+
+       //call @matmul1(%A, %B, %C1) : (memref<?x?xf64>, memref<?x?xf64>, memref<?x?xf64>) -> ()
```
Origami404, line 258: I would recommend splitting these functions into different files, so it will be way more convenience to test different situations using different make targets, instead of manually changing the comment here.

hxdtest: OK, I will split these function into different files.

examples/MLIRGEMM/gemm.mlir
hxdtest: > @hxdtest Thanks for providing the examples! (1) Please reformat your code as @meshtag said above. (2) How do you measure the execution time? Can we add a GFLOPS report to the example? (see [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/benchmark/conv2d-buddy-template#L69) for an example) (3) Have you considered adding a vectorization step to the example?

（1）I am confused about the format. Do we have some scripts like `pre-commit run` to check and reformat the code? 
（2）The GFLOPS report will be added in next commits.
  (3)   vectorization step will be added in next commits.

Origami404: > （1）I am confused about the format. Do we have some scripts like pre-commit run to check and reformat the code?

In my understanding, what @meshtag suggested is a changing of proper PR title and commit message. For example, you may change your commit message into something like `[examples][GEMM] add GEMM example`, using brackets to pointed out what part these commits are in. You may also provide more details in commit message like `add GEMM with different implementation` then lists them.

For code style in examples, we looks like don't have any automatically check yet, due to the lack of MLIR formatting tools. Code style checking of examples are usually done in PR manually. Your code style looks good for me for most of them, but I do notice some problems:

1. Unify the indents. Some of them are 4 spaces and others are 2, that's not good.
2. Giving reasons about these newlines. Normally one newline is enough, if in somewhere you use 2 or more of them, you may ask yourself "Why there is 2 or more newline?"
3. Files should be ended by a LF character.

BTW, I strongly suggest you to post this comment in PR rather than in my review comments, so more people may notice this. And the review comments will be folded after review is done, so it should only contains information related to the review only.

zhanghb97:
@hxdtest The recommended way to format the MLIR file is to use mlir-opt, for example:
```
$ mlir-opt gemm.mlir -o formatted-gemm.mlir
```

examples/MLIRGEMM/makefile
```diff
@@ -0,0 +1,41 @@
...
+		-convert-func-to-llvm -reconcile-unrealized-casts | \
+	${MLIR_TRANSLATE} --mlir-to-llvmir -o log.ll
+
+gemm-run:
+	@${MLIR_OPT} ./gemm.mlir ${MLIR_OPT_OPTIONS} \
+		-convert-linalg-to-loops -lower-affine -convert-scf-to-cf \
+		-convert-vector-to-llvm -convert-memref-to-llvm -convert-arith-to-llvm \
+		-convert-func-to-llvm -reconcile-unrealized-casts | \
+	${MLIR_CPU_RUNNER} ${OPT_FLAG} -e main -entry-point-result=void -shared-libs=${MLIR_RUNNER_UTILS} -shared-libs=${MLIR_C_RUNNER_UTILS}
+
```
zhanghb97, line 41: As the warning here, please add a new line at the end of the file.

examples/MLIRGEMM/gemm.mlir
hxdtest: This example is removed.

examples/MLIRGEMM/gemm.mlir
hxdtest: Many thanks for detailed explanation. PR title is modified as you suggested and code style is fixed.

examples/MLIRGEMM/tiling-packing-vector-matmul.mlir
```diff
@@ -0,0 +1,166 @@
...
+    outs(%C:memref<?x?xf64>)
+
+
+    // Execution times.
+    %reps = arith.constant 1 : index
+
+    // Record start time.
+    %t_start = call @rtclock() : () -> f64  
+
+    // Execute convolution for specific times.
```
xlinsist, line 133: Change convolution -> matmul. Other .mlir files ditto.

PR #125, opened 2023-03-07
[RVV] Extend operands' data type in RVV dialect to support rv32

xlinsist:
To attain RVV intrinsics in rv32, for current operations in RVV dialects, change the data type of `vl` from `Index` to `AnySignlessIntegerOrIndex`. Here, `vl` is an essential parameter for RVV to represent current vector length to process.

After changing:
1. Both  `%vl = rvv.setvl %avl, %sew, %lmul : Index` and `%vl = rvv.setvl %avl, %sew, %lmul : i32` can be lowered correctly, which means the usage of `rvv.setvl` in original programs will not be effected.
2. Both Index integers and i32 integers can be passed into RVV operations as parameters. When passing i32 integers, the generated RVV intrinsics will be of rv32 instead of rv64.

To test the correctness,  a **i32-type matmul demo(vector-exp-predication-matmul-i32.mlir)** is added in /buddy-mlir/examples/VectorExpDialect, from which rv32 asemble code can be generated with these modifications in RVV dialect.

### DIFF
--- a/midend/lib/Dialect/RVV/Transforms/LegalizeForLLVMExport.cpp
+++ b/midend/lib/Dialect/RVV/Transforms/LegalizeForLLVMExport.cpp
@@ -118,7 +118,7 @@ struct RVVLoadOpLowering : public ConvertOpToLLVMPattern<RVVLoadOp> {
     Value vl = loadOp.getOperand(2);
     Value vlCast = rewriter
                        .create<UnrealizedConversionCastOp>(
-                           loadOp.getLoc(), rewriter.getI64Type(), vl)
+                           loadOp.getLoc(), rewriter.getI32Type(), vl)
                        .getResult(0);
     rewriter.replaceOpWithNewOp<RVVIntrLoadEleOp>(loadOp, resultType, passthru,
                                                   bitCastedPtr, vlCast);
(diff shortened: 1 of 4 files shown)


midend/lib/Dialect/RVV/Transforms/LegalizeForLLVMExport.cpp
```diff
@@ -118,7 +118,7 @@ struct RVVLoadOpLowering : public ConvertOpToLLVMPattern<RVVLoadOp> {
     Value vl = loadOp.getOperand(2);
     Value vlCast = rewriter
                        .create<UnrealizedConversionCastOp>(
-                           loadOp.getLoc(), rewriter.getI64Type(), vl)
+                           loadOp.getLoc(), rewriter.getI32Type(), vl)
```
xlinsist, line 121: Note that this modification explictly consider `vl` as i32 type, even in rv64 environment. It will not effect the correctness of original programs because of upward compatibility(i.e. rv32 intrinsics can run on rv64), but it should be better to get integer type based on platform-specific bit width. @zhanghb97

xlinsist:
[Relate commit](https://github.com/buddy-compiler/buddy-mlir/commit/cc57e60f073d769224a334b4df10c16b43eb2711) has been pushed to upstream. Closing it.

PR #127, opened 2023-03-15
[DIP] accelerate rotation2D

FlagerLee:
rotation2D is now accelerated using affine transform and parallel(openmp).

Main contribution are as follows:

1. Rewrite DIPResize2DOpLowering using affine transform(the same as opencv), which can also be used on resize2D. It also reduces unnecessary ops.
2. Avoid using function IotaVec. IotaVec uses lots of load-store operations which may result in reduced performance. I used a method to circumvent it. First, given a vector initialized to (0,1,... ,stride-1). Then add a multiple of 'stride' to this vector in each loop body (since the step of the inner loop is equal to stride, just add the inner loop variable value directly).
3. Use parallel to accelerate it. The perfect nested loop can be paralized perfectly. I replaced buildAffineLoopNest using scf.parallel, and lower this operation to openmp. The time cost of this approach is 1/3 of the original(1/2 when using async).

However, openmp must be enabled when building llvm, because I'm using clang++ to compile rotation2D.cpp.


zhanghb97:
@FlagerLee Nice!
Yuchen did a great optimization with multithreading! 
@meshtag Please have a look!

meshtag:
Thanks a lot for the PR @FlagerLee !
I am not getting much time due to my university exams this week, I'll have a look as soon as I can.

frontend/Interfaces/lib/CMakeLists.txt
```diff
@@ -1,35 +1,37 @@
...
-elseif(HAVE_SSE)
-  set(SPLITING_SIZE 64)
-elseif(HAVE_NEON)
-  set(SPLITING_SIZE 64)
-endif()
+    set(SPLITING_SIZE ${BUDDY_DIP_OPT_STRIP_MINING})
+elseif (HAVE_AVX512)
+    set(SPLITING_SIZE 256)
+elseif (HAVE_AVX2)
+    set(SPLITING_SIZE 16)
```
meshtag, line 6: Did you compare performance between splitting size 16 and 128?

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
   return builder.create<arith::DivFOp>(loc, sinVal, cosVal);
 }
 
+SmallVector<Value, 6> getRotationMatrix(OpBuilder& builder, Location loc, Value centerX, Value centerY, Value angle, Value scale) {
```
meshtag, line 414: Please add a comment explaining the use of this function here as well. Refer other functions in this file.

midend/include/Utils/DIPUtils.h
```diff
@@ -111,6 +111,13 @@ void fillPixels(OpBuilder &builder, Location loc, Value resXVec, Value resYVec,
 // Calculate tan(angle / 2) where angle is a function parameter.
 Value customTanVal(OpBuilder &builder, Location loc, Value angleVal);
 
+SmallVector<Value, 6> getRotationMatrix(OpBuilder& builder, Location loc, Value centerX, Value centerY, Value angle, Value scale);
```
meshtag, line 114: Add a small comment explaining the use of this function for consistency.

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+  Value m1 = beta;
+  Value m2 = builder.create<arith::SubFOp>(loc, m20, m21);
+  Value m3 = builder.create<arith::NegFOp>(loc, beta);
+  Value m4 = alpha;
+  Value m5 = builder.create<arith::AddFOp>(loc, m50, m51);
+
+  return SmallVector<Value, 6>{m0, m1, m2, m3, m4, m5};
+}
+
+inline void inverseAffineMatrix(OpBuilder &builder, Location loc,
```
meshtag, line 435: Same for here as well.

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+  Value m2pm3 =
+      builder.create<arith::MulFOp>(loc, affineMatrix[2], affineMatrix[3]);
+  Value m4pm5 =
+      builder.create<arith::MulFOp>(loc, affineMatrix[4], affineMatrix[5]);
+  Value negB2 = builder.create<arith::AddFOp>(loc, m2pm3, m4pm5);
+  affineMatrix[2] = builder.create<arith::NegFOp>(loc, negB1);
+  affineMatrix[5] = builder.create<arith::NegFOp>(loc, negB2);
+}
+
+void affineTransformController(OpBuilder &builder, Location loc,
```
meshtag, line 480: Same for here as well.

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+  Value inputRow = builder.create<memref::DimOp>(loc, input, c0);
+  Value inputCol = builder.create<memref::DimOp>(loc, input, c1);
+
+  Value outputRow = builder.create<memref::DimOp>(loc, output, c0);
+  Value outputCol = builder.create<memref::DimOp>(loc, output, c1);
+
+  Value strideVal = builder.create<arith::ConstantIndexOp>(loc, stride);
+
+  // generate vector [0., 1., ..., stride - 1]
+  std::vector<float> xInit(stride);
```
meshtag, line 520: Can you consider adding a generic function for this in `Utils.cpp`? ([Ref](https://github.com/buddy-compiler/buddy-mlir/blob/cc57e60f073d769224a334b4df10c16b43eb2711/midend/lib/Utils/Utils.cpp#L102))

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+      /*upperBounds=*/ValueRange{outputRow, outputCol},
+      /*steps=*/ValueRange{c1, strideVal},
+      /*bodyBuilderFn=*/
+      [&](OpBuilder &builderFn, Location locFn, ValueRange ivsFn) {
+        Value yVec = builderFn.create<vector::SplatOp>(
+            locFn, vectorTyF32, indexToF32(builderFn, locFn, ivsFn[0]));
+        Value delta = builderFn.create<vector::SplatOp>(locFn, vectorTyF32, indexToF32(builderFn, locFn, ivsFn[1]));
+        Value xVec = builderFn.create<arith::AddFOp>(locFn, xVecInitial, delta);
+
+        // sourceXVec=(xVec*m0Vec*1024+(yVec*m1Vec+m2Vec)*1024+512)/1024
```
meshtag, line 539: Please replace 1024 and 512 with appropriate image dimension variables, since this is not the general case and we intend to work with images of different dimensions.

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+            [&](OpBuilder &builderFor, Location locFor, ValueRange ivsFor,
+                ValueRange iterArg) {
+              Value inputPixelX = builderFor.create<vector::ExtractElementOp>(locFor, inputXVec, ivsFor[0]);
+              Value inputPixelY = builderFor.create<vector::ExtractElementOp>(locFor, inputYVec, ivsFor[0]);
+              Value inputPixelXI = F32ToIndex(builderFor, locFor,
+                                              builderFor.create<math::RoundOp>(locFor, inputPixelX));
+              Value inputPixelYI = F32ToIndex(builderFor, locFor,
+                                              builderFor.create<math::RoundOp>(locFor, inputPixelY));
+
+              auto inBound = [&](Value lb, Value ub, Value val) {
```
meshtag, line 569: Perhaps consider using [this generic](https://github.com/buddy-compiler/buddy-mlir/blob/cc57e60f073d769224a334b4df10c16b43eb2711/midend/lib/Utils/Utils.cpp#L95) function? Feel free to propose a change in its implementation if you want to.

meshtag (changes requested):
I have suggested some changes. I'll build and test your branch once they are addressed. 
Can you please provide me a link to the benchmark branch, which you used to get performance data for this implementation? 

This is a nice addition overall, thanks!

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
   return builder.create<arith::DivFOp>(loc, sinVal, cosVal);
 }
 
+SmallVector<Value, 6> getRotationMatrix(OpBuilder& builder, Location loc, Value centerX, Value centerY, Value angle, Value scale) {
+  Value alpha0 = builder.create<math::CosOp>(loc, angle);
+  Value alpha = builder.create<arith::MulFOp>(loc, alpha0, scale);
+  Value beta0 = builder.create<math::SinOp>(loc, angle);
+  Value beta = builder.create<arith::MulFOp>(loc, beta0, scale);
+  Value oneMinusAlpha = builder.create<arith::SubFOp>(loc, builder.create<arith::ConstantOp>(loc, builder.getF32FloatAttr((float)1.)), alpha);
+  SmallVector<Value, 6> mat;
```
meshtag, line 420: This one as well variable declared on lines 425, 426 and 429 seem redundant, are they not?

meshtag, line 420: This one as well variables declared on lines 425, 426 and 429 seem redundant, are they not?

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+inline void inverseAffineMatrix(OpBuilder &builder, Location loc,
+                                SmallVector<Value, 6> &affineMatrix) {
+  Value c0F32 = builder.create<arith::ConstantOp>(
+      loc, builder.getF32FloatAttr((float).0));
+  Value m0pm4 =
+      builder.create<arith::MulFOp>(loc, affineMatrix[0], affineMatrix[4]);
+  Value m1pm3 =
+      builder.create<arith::MulFOp>(loc, affineMatrix[1], affineMatrix[3]);
+  Value D = builder.create<arith::SubFOp>(loc, m0pm4, m1pm3);
+  // TODO: can a float number compare with 0.0 directly?
```
meshtag, line 444: What exactly is the intent behind this comment?

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+  Value dEq0 =
+      builder.create<arith::CmpFOp>(loc, arith::CmpFPredicate::OEQ, D, c0F32);
+  auto scfRes = builder.create<scf::IfOp>(
```
meshtag, line 447: Can you modify the condition such that we do not require the `else` block at all?

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+        Value c1F32 = elseBuilder.create<arith::ConstantOp>(
+            elseLoc, builder.getF32FloatAttr((float)1.));
+        Value res = elseBuilder.create<arith::DivFOp>(elseLoc, c1F32, D);
+        elseBuilder.create<scf::YieldOp>(elseLoc, ValueRange{res});
+      });
+  D = scfRes.getResult(0);
+  Value negD = builder.create<arith::NegFOp>(loc, D);
+  Value A11 = builder.create<arith::MulFOp>(loc, affineMatrix[4], D);
+  Value A22 = builder.create<arith::MulFOp>(loc, affineMatrix[0], D);
+  affineMatrix[0] = A11;
```
meshtag, line 462: Why not directly assign the operation result itself, instead of introducing `A11`?
Same argument for `A22` as well.

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +411,185 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+
+void affineTransformController(OpBuilder &builder, Location loc,
+                               MLIRContext *ctx, Value input, Value output,
+                               SmallVector<Value, 6> affineMatrix,
+                               int64_t stride) {
+  VectorType vectorTyF32 = VectorType::get({stride}, FloatType::getF32(ctx));
+
+  Value c0 = builder.create<arith::ConstantIndexOp>(loc, 0);
+  Value c1 = builder.create<arith::ConstantIndexOp>(loc, 1);
+  Value c512F32 =
```
meshtag, line 488: What is the significance of 512 and 1024 here? Will these constants cause any loss of generality in the design?

meshtag (changes requested):

frontend/Interfaces/lib/CMakeLists.txt
zhanghb97: We may encounter such a problem that optimization algorithms for different ops seem to need different configurations.

FlagerLee:
I'm busying with my homework now. I'll have a look next week.

frontend/Interfaces/lib/CMakeLists.txt
FlagerLee: I have compared performance on my own machine(with avx2) using stride = 2^k(k in [1, 8]). The best performance appears when stride=16 and 32. As Hongbin said, the best parameter differs on different machine, so I can only promise that this parameter is the best on _my_ machine.

midend/lib/Utils/DIPUtils.cpp
FlagerLee: this method cannot be used if the start index is a Value. I initialize it using std::vector<float>, not std::vecctor<Value>

midend/lib/Utils/DIPUtils.cpp
FlagerLee: Put these into configuration?

midend/lib/Utils/DIPUtils.cpp
FlagerLee: Function ValBound returns a Value, while this function returns boolean.
ValBound: return 0 if val < 0, val if 0 <= val <= lastElemf32, lastElemf32 if lastElemf32 < val
inBound: return true if lb <= val <= ub, false if val < lb or ub < val

midend/lib/Utils/DIPUtils.cpp
FlagerLee: This is only written to make it easier to understand the code, and does not generate redundant statements

midend/lib/Utils/DIPUtils.cpp
FlagerLee: I don't know what will happen in mlir if I wrote a line of code: if(x == 0.0). If I needed to do a comparison of a floating point number to a zero using C, I would write if(fabs(x) < 1e-6). So I'm not sure if this **is** absolutely correct.

midend/lib/Utils/DIPUtils.cpp
FlagerLee: No. 0 must not be divided.

frontend/Interfaces/lib/CMakeLists.txt
meshtag: > optimization algorithms for different ops seem to need different configurations.

Is this behaviour dependent on the algorithm or the problem size itself (Image dimensions in this case) ?

midend/lib/Utils/DIPUtils.cpp
meshtag: > Put these into configuration?

Can you elaborate.

midend/lib/Utils/DIPUtils.cpp
meshtag: I think the intent at both places is similar i.e ensuring that the estimated pixel co-ordinates do not exceed output image dimensions. The function which you wrote does this by filtering pixels who have their estimated co-ordinates outside the output image, while `valBound` does this by clamping the dimension exceeding values to minimal/maximal allowed output image dimension value.

midend/lib/Utils/DIPUtils.cpp
meshtag: Perhaps consider using the "Not Equal" (ne mnemonic) [here in the predicate](https://github.com/FlagerLee/buddy-mlir/blob/3a280a09d72c82aa1d2c9c734e250fb6e42586e1/midend/lib/Utils/DIPUtils.cpp#L446)? 
(Ref: [attribute definitions for determining the type of comparison](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithcmpi-mlirarithcmpiop))

midend/lib/Utils/DIPUtils.cpp
meshtag: > if(x == 0.0)

This will not generate comparison code. If you wish to generate IR that does comparisons at runtime, you need to use `cmp` ops. Lowering pass generates lowered IR at compile time which can then be used to form an executable.

frontend/Interfaces/lib/CMakeLists.txt
FlagerLee: No. I have discussed with Hongbin on this issue and concluded that this parameter is to fully utilize the cpu's simd registers.

midend/lib/Utils/DIPUtils.cpp
FlagerLee: Okay. I'll add this in Utils.cpp.

midend/lib/Utils/DIPUtils.cpp
FlagerLee: 1024 and 512 are not dimension-based parameters. It depends on how much precision you need when converting a float to an int. Of course, it seems to have other factors to consider that also come into play in Opencv's remap function. I have plans for the next week to put together the full flow of Opencv rotations.

midend/lib/Utils/DIPUtils.cpp
FlagerLee: Yes, but these two functions are not interchangeable

midend/lib/Utils/DIPUtils.cpp
FlagerLee: This comment is for line 456 and 457

midend/lib/Utils/DIPUtils.cpp
FlagerLee: Not the problem you considered. If scf.if op produces results, 'else' region must have exactly one block. You can check the [documentation](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfif-mlirscfifop).

midend/lib/Utils/DIPUtils.cpp
meshtag: Ahh, I see

midend/lib/Utils/DIPUtils.cpp
meshtag: Yes, but can you achieve the same functionality from both of them? 

Again, my purpose here is code reuse since pixel bounding is also used in resizing. If you think you need to change the implementation of `valBound` itself, feel free to propose a change and I'll review it.

midend/lib/Utils/DIPUtils.cpp
meshtag: Then perhaps put this comment at that location? or make this intent more clear by mentioning this in the comment itself?

FlagerLee: Okay

FlagerLee:
This branch current removed multithread because it makes it difficult to continue the optimization of single threads. Multithread would be added when the single thread performance is high enough.

FlagerLee:
@meshtag

midend/lib/Utils/DIPUtils.cpp
meshtag: From what I see, you are not using `mat` anywhere in this function itself. In my opinion, we should avoid such redundancy.

frontend/Interfaces/lib/CMakeLists.txt
```diff
@@ -1,35 +1,37 @@
...
-              -o ${CMAKE_CURRENT_BINARY_DIR}/DIP.o
-  DEPENDS buddy-opt
-  )
+        POST_BUILD
+        COMMAND ${BUDDY_BINARY_DIR}/buddy-opt ${CMAKE_CURRENT_SOURCE_DIR}/DIP.mlir
+        -lower-dip="DIP-strip-mining=${SPLITING_SIZE}"
+        -arith-expand
+        -lower-affine
+        -convert-scf-to-openmp
+        -convert-async-to-llvm
```
meshtag, line 20: I don't think you require this pass if you are removing multi-threading support.

README.md
```diff
@@ -25,7 +25,7 @@ $ cd buddy-mlir
 $ mkdir llvm/build
 $ cd llvm/build
 $ cmake -G Ninja ../llvm \
-    -DLLVM_ENABLE_PROJECTS="mlir;clang" \
+    -DLLVM_ENABLE_PROJECTS="mlir;clang;openmp" \
```
meshtag, line 28: I don't think this change is required now.

examples/DIPDialect/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ if(BUDDY_ENABLE_OPENCV)
   include_directories(${OpenCV_INCLUDE_DIRS})
 endif()
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp -no-pie")
```
meshtag, line 6: I don't think this change is required now.

meshtag (changes requested):
Thanks @FlagerLee! 
I have added some comments, we should try to get maximal performance via single threading in this patch, and then we can deal with multi-threading stuff in a separate patch.

FlagerLee:
The best performance still needs several weeks. It's hard.

meshtag:
@FlagerLee , what is the status of this PR? 
Can you please resolve merge conflicts and attach the benchmark results.

FlagerLee:
This failure is weird, I've seen this when writing tests. It seems that it might failed randomly. I'll find it out.

FlagerLee:
@meshtag conflict resolved

meshtag:
A buddy-benchmark PR would be really great, I think. This would help us keep a track of the performance as this change is about "acceleration". I just want to log our current state of performance so that we can improve it from there (if we can).

FlagerLee:
Can I just show you here now? Cause it takes time to write benchmark.

meshtag:
Okay, that works for now.

FlagerLee:
Currently I cannot give you **real** execution time using buddy-benchmark. Reasons are:
* Extra Constructor [`Img<float, 2> inputBuddyRotate2D(inputImageBuddyRotate2D);` ](https://github.com/taiqzheng/buddy-benchmark/blob/rotate_dev/benchmarks/ImageProcessing/BuddyRotate2DBenchmark.cpp#LL77C3-L77C61)
* Awful implementation in dip::Rotate2D. A new Memref entity is created in its implementation, and each time a Memref entity is created all the memory in it is set to 0 (instead of the uninitialized state). Also, it returns a Memref **Value** but not a pointer or constant reference or etc., will result in a call to the copy constructor and spent large amont of time.

When I said 'My implementation is faster than opencv's', it usually indicates that dip::detail::_mlir_ciface_rotate_2d is faster than cv::warpAffine, but not dip::Rotate2D is faster than cv::warpAffine.

I'm writing shell script to show you the speed difference between dip::detail::_mlir_ciface_rotate_2d and cv::warpAffine.

FlagerLee:
function [test_implementation]() using _mlir_ciface_rotate_2d:
```
Mat image = imread(argv[1], IMREAD_GRAYSCALE);
if (image.empty()) {
  cout << "Could not read the image: " << argv[1] << endl;
}

// Define memref containers.
Img<float, 2> input(image);
float angle = 120;
float angleRad = M_PI * angle / 180;

float sinAngle = std::sin(angleRad);
float cosAngle = std::cos(angleRad);

int outputRows = std::round(std::abs(input.getSizes()[0] * cosAngle) +
                            std::abs(input.getSizes()[1] * sinAngle)) +
                 1;
int outputCols = std::round(std::abs(input.getSizes()[1] * cosAngle) +
                            std::abs(input.getSizes()[0] * sinAngle)) +
                 1;

intptr_t sizesOutput[2] = {outputRows, outputCols};
MemRef<float, 2> output(sizesOutput);

auto start = system_clock::now();
dip::detail::_mlir_ciface_rotate_2d(&input, angleRad, &output);
auto end = system_clock::now();
auto duration = duration_cast<microseconds>(end - start);

cout << double(duration.count()) * microseconds::period::num / microseconds::period::den << endl;

// Define a cv::Mat with the output of Rotate2D.
Mat outputImageRotate2D(output.getSizes()[0], output.getSizes()[1], CV_32FC1,
                        output.getData());
imwrite(argv[2], outputImageRotate2D);

return 1;
```

function test_implementation using cv::warpAffine:
```
Mat image = imread(argv[1], IMREAD_GRAYSCALE);
if (image.empty()) {
  cout << "Could not read the image: " << argv[1] << endl;
}

setNumThreads(0);

int width = image.cols;
int height = image.rows;

float angle = 120;

double radian= angle * CV_PI / 180.;
double width_rotate = fabs(width*cos(radian))+fabs(height*sin(radian));
double height_rotate= fabs(width*sin(radian)) + fabs(height*cos(radian));

Point2f center((float)width / 2.0, (float)height/ 2.0);
Mat m1 = cv::getRotationMatrix2D(center, angle, 1.0);
m1.at<double>(0, 2) += (width_rotate - width) / 2.;
m1.at<double>(1, 2) += (height_rotate - height) / 2.;
Mat imgOut;

auto start = system_clock::now();
warpAffine(image, imgOut, m1, cv::Size(width_rotate, height_rotate), cv::INTER_NEAREST);
auto end = system_clock::now();
auto duration = duration_cast<microseconds>(end - start);


cout << double(duration.count()) * microseconds::period::num / microseconds::period::den << endl;
imwrite(argv[2], imgOut);

return 1;
```
Notice that I have set cv::numThreads to 0 so that opencv won't use multithread to accelerate cv::warpAffine.

I then run these two program 10 times each and calculate total time, the result is:
buddy time is: .749989
opencv time is: .746589
(opencv performs 0.01s faster than before)

frontend/Interfaces/lib/CMakeLists.txt
```diff
@@ -1,39 +1,43 @@
...
 
 add_library(BuddyLibDIP STATIC DIP.o)
 
 SET_TARGET_PROPERTIES(BuddyLibDIP PROPERTIES
-  LINKER_LANGUAGE C
-  ARCHIVE_OUTPUT_DIRECTORY ${LIBRARY_OUTPUT_DIRECTORY}
-  )
+        LINKER_LANGUAGE C
+        ARCHIVE_OUTPUT_DIRECTORY ${LIBRARY_OUTPUT_DIRECTORY}
+        )
```
meshtag, line 43: Nit: Add an extra line here.

frontend/Interfaces/lib/CMakeLists.txt
```diff
@@ -1,39 +1,43 @@
...
-              --filetype=obj
-              -o ${CMAKE_CURRENT_BINARY_DIR}/DIP.o
-  DEPENDS buddy-opt
-  )
+        POST_BUILD
+        COMMAND ${CMAKE_BINARY_DIR}/bin/buddy-opt ${CMAKE_CURRENT_SOURCE_DIR}/DIP.mlir
+        -lower-dip="DIP-strip-mining=${SPLITING_SIZE}"
+        -arith-expand
+        -lower-affine
+        #-convert-scf-to-openmp
```
meshtag, line 20: Nit: If you are not using these passes, just remove them imo.

frontend/Interfaces/lib/CMakeLists.txt
```diff
@@ -1,39 +1,43 @@
...
-elseif(HAVE_AVX2)
-  set(SPLITING_SIZE 128)
-elseif(HAVE_SSE)
-  set(SPLITING_SIZE 64)
-elseif(HAVE_NEON)
-  set(SPLITING_SIZE 64)
-endif()
+    set(SPLITING_SIZE ${BUDDY_DIP_OPT_STRIP_MINING})
+elseif (HAVE_AVX512)
+    # TODO: current rotation generates wrong image using avx512
```
meshtag, line 4: What do you mean by wrong? Can you please elaborate? Also, please link corresponding github issue here.

midend/include/Utils/AffineTransformUtils.h
```diff
@@ -0,0 +1,37 @@
+//
+// Created by flagerlee on 4/27/23.
```
meshtag, line 2: Please add the license after having a look at other files in this directory.

midend/lib/Utils/AffineTransformUtils.cpp
```diff
@@ -0,0 +1,203 @@
+//
+// Created by flagerlee on 4/27/23.
```
meshtag, line 2: Add the license here as well.

midend/lib/Utils/AffineTransformUtils.cpp
```diff
@@ -0,0 +1,203 @@
...
+                    ifBuilder.create<scf::YieldOp>(ifLoc);
+                  });
+
+              xBuilder.create<scf::YieldOp>(xLoc);
+            });
+
+        yBuilder.create<scf::YieldOp>(yLoc);
+      });
+}
+} // namespace buddy
```
meshtag, line 203: Add a new line here

midend/lib/Utils/DIPUtils.cpp
meshtag: Is this resolved?

midend/lib/Utils/DIPUtils.cpp
meshtag: I'd prefer having the code in the "if" block, you can keep the "else" block empty

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +412,186 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+  Value strideVal = builder.create<arith::ConstantIndexOp>(loc, stride);
+  Value outputColStrideRatio =
+      builder.create<arith::DivUIOp>(loc, outputCol, strideVal);
+  Value outputColMultiple = builder.create<arith::MulIOp>(
+      loc, builder.create<arith::AddIOp>(loc, outputColStrideRatio, c1Index),
+      strideVal);
+
+  // generate vector [0., 1., ..., stride - 1]
+  Value xVecInitial = iotaVec0F32(builder, loc, stride);
+  // Value xVecInitial = builder.create<vector::SplatOp>(loc, vectorTyF32,
```
meshtag, line 531: Please delete this, commenting it here doesn't make much sense to me.

meshtag (changes requested):

meshtag:
@FlagerLee, are you able to build `rotation2D` target on this branch? I tried building it on my system and am getting some errors, I suspect these are arising due to incorrect merge with upstream main.

FlagerLee:
> @FlagerLee, are you able to build `rotation2D` target on this branch? I tried building it on my system and am getting some errors, I suspect these are arising due to incorrect merge with upstream main.

I'll try it.

frontend/Interfaces/lib/CMakeLists.txt
FlagerLee: The output of the picture is wrong. It only happens in new code. I'll open an issue to solve this problem.

meshtag:
> Awful implementation in dip::Rotate2D. A new Memref entity is created in its implementation, and each time a Memref entity is created all the memory in it is set to 0 (instead of the uninitialized state). Also, it returns a Memref Value but not a pointer or constant reference or etc., will result in a call to the copy constructor and spent large amont of time.

The reason behind the slightly different interface in `rotate_2d` (as compared to correlation APIs) is to save the user from the hassle of creating a container with correct size for the rotation output. 

As of now, the user just needs to pass the input, angle of rotation and angle type in the API. If I were to change this, they would also have to construct the container like this 
```
int outputRows = std::round(std::abs(input->getSizes()[0] * cosAngle) +
                              std::abs(input->getSizes()[1] * sinAngle)) +
                   1;
  int outputCols = std::round(std::abs(input->getSizes()[1] * cosAngle) +
                              std::abs(input->getSizes()[0] * sinAngle)) +
                   1;

```

If you think that the former is a better design from user perspective as well, I can change the interface. Let me know you opinion on this.

frontend/Interfaces/lib/CMakeLists.txt
meshtag: Does it work properly on other backends? (AVX2, NEON)

FlagerLee:
We are planning to reconstruct Img container. We may modify these Interfaces months later.

frontend/Interfaces/lib/CMakeLists.txt
FlagerLee: AVX2 works properly. I don't have NEON machine.

midend/lib/Utils/DIPUtils.cpp
FlagerLee: I don't understand why can I keep else block empty. The origin code is
```
D == 0. ? 1./D : D;
```

FlagerLee: > I'd prefer having the code in the "if" block, you can keep the "else" block empty

I don't understand why can I keep else block empty. The origin code is
```
D == 0. ? 1./D : D;
```

frontend/Interfaces/lib/CMakeLists.txt
meshtag: This is strange behaviour, I would prefer accuracy over most things. Did you investigate the reason behind wrong image output on AVX 512?

meshtag:
FYI, I am getting the following error when I try to run the example 
```
root@3eb35749d59c:~/forks/flaggerlee_fork/buddy-mlir/build/bin# ./rotation2D ../../examples/images/YuTu.png result-dip-rotate.png
free(): invalid next size (normal)
Aborted (core dumped)
```

Please let me know if this is reproducible at your end.

FlagerLee:
@meshtag Problem resolved.

frontend/Interfaces/lib/CMakeLists.txt
FlagerLee: Bug fixed.

FlagerLee:
@zhanghb97 Request for review

midend/lib/Utils/Utils.cpp
zhanghb97, line 1: Please format this file with clang-format.

midend/lib/Conversion/LowerDIP/LowerDIPPass.cpp
zhanghb97, line 1: Please format this file with clang-format.

midend/include/Utils/Utils.h
zhanghb97, line 1: Please format this file with clang-format.

midend/include/Utils/DIPUtils.h
zhanghb97, line 1: Please format this file with clang-format.

midend/lib/Conversion/LowerDIP/LowerDIPPass.cpp
```diff
@@ -187,142 +187,43 @@ class DIPRotate2DOpLowering : public OpRewritePattern<dip::Rotate2DOp> {
...
+    // let alpha = scale * cos(angle), beta = scale * sin(angle)
+    // the affine matrix would be as follow:
+    // [[alpha, beta, (1 - alpha) * centerx - beta * centery],
+    //  [-beta, alpha, beta * centerx + (1 - alpha) * centery]]
+    Value centerX = rewriter.create<arith::ShRSIOp>(loc, inputCol, c1);
+    Value centerY = rewriter.create<arith::ShRSIOp>(loc, inputRow, c1);
+    Value centerXF32 = indexToF32(rewriter, loc, centerX);
+    Value centerYF32 = indexToF32(rewriter, loc, centerY);
+
+    auto affineMatrix = dip::getRotationMatrix(rewriter, loc, centerXF32, centerYF32, angleVal, c1F32);
```
zhanghb97, line 213: It seems only f32 inputs are currently supported. Will we support all types of input images here? If we're going to start with just f32 for now, maybe we should add a TODO comment stating so!

midend/lib/Utils/DIPUtils.cpp
```diff
@@ -410,6 +412,182 @@ Value customTanVal(OpBuilder &builder, Location loc, Value angleVal) {
...
+  MemRefType dynamicTypeI32 =
+      MemRefType::get(ShapedType::kDynamic, IntegerType::get(ctx, 32));
+
+  // compute x*m0+m2 and x*m3+m5 and store the results into xMm0 and xMm3
+  Value xMm0 =
+      builder.create<memref::AllocOp>(loc, dynamicTypeI32, outputColMultiple);
+  Value xMm3 =
+      builder.create<memref::AllocOp>(loc, dynamicTypeI32, outputColMultiple);
+
+  const int RSV_BITS = 5;
```
zhanghb97, line 539: There is a missing explanation of the RSV_BITS variable and why it is assigned a value of 5. Please add some comments here.

frontend/Interfaces/lib/CMakeLists.txt
```diff
@@ -1,35 +1,35 @@
...
-              -reconcile-unrealized-casts | 
-          ${LLVM_MLIR_BINARY_DIR}/mlir-translate --mlir-to-llvmir |
-          ${LLVM_MLIR_BINARY_DIR}/llc 
-              -mtriple=${BUDDY_TARGET_TRIPLE}
-              -mattr=${BUDDY_OPT_ATTR}
-              --filetype=obj
-              -o ${CMAKE_CURRENT_BINARY_DIR}/DIP.o
-  DEPENDS buddy-opt
-  )
+        POST_BUILD
```
zhanghb97, line 14: Why should we use POST_BUILD here?

zhanghb97 (changes requested):
LGTM, with a few minor changes.

zhanghb97 (approved):
Thanks!

PR #135, opened 2023-04-30
Add support for lld linker in build config of llvm

Ris-Bali:
By default we use the gnu linker to build llvm. lld is a linker from the llvm project and is much [faster](https://llvm.org/devmtg/2017-10/slides/Ueyama-lld.pdf) than the default gnu linker. This PR aims to give the user an option to use the lld linker while building llvm.


zhanghb97:
Thanks!

PR #145, opened 2023-05-15
[examples][MLIRGPU] Add two applications related to GPU dialect 

buaayxy:
Add two applications, explains how to use shared memory, and create a pipeline to transform affine dialect to gpu dialect

### DIFF
--- a/examples/MLIRGPU/affine-to-gpu.mlir
+++ b/examples/MLIRGPU/affine-to-gpu.mlir
@@ -1,0 +1,32 @@
+module {
+func.func private @printMemrefF32(memref<*xf32>)
+
+func.func @main() {
+    %a = memref.alloc() : memref<8192xf32>
+    %b = memref.alloc() : memref<8192xf32>
+    %c = memref.alloc() : memref<8192xf32>
+    %cf1 = arith.constant 1.0 : f32
+    %cf2 = arith.constant 2.0 : f32
+    linalg.fill
+    ins(%cf1: f32)
+    outs(%a: memref<8192xf32>)
+    linalg.fill
+    ins(%cf2: f32)
+    outs(%b: memref<8192xf32>)
+    %cast_a = memref.cast %a : memref<8192xf32> to memref<*xf32>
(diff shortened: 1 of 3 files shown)


examples/MLIRGPU/affine-to-gpu.mlir
```diff
@@ -0,0 +1,32 @@
...
+        %1 = affine.load %a[%idx0] : memref<8192xf32>
+        %2 = affine.load %b[%idx0] : memref<8192xf32>
+        %3 = arith.addf %1, %2 : f32
+        affine.store %3, %c[%idx0] : memref<8192xf32>
+    }
+    // %print_out = memref.cast %c : memref<8192xf32> to memref<*xf32>
+    func.call @printMemrefF32(%cast_c) : (memref<*xf32>) -> ()
+    func.return
+}
+}
```
linuxlonelyeagle, line 32: Add endline.

examples/MLIRGPU/gpu-matmul.mlir
```diff
@@ -0,0 +1,84 @@
+module attributes {gpu.container_module} {
+  gpu.module @kernels {
+    gpu.func @mutmal(%matA: memref<1024x1024xf32>, %matB: memref<1024x1024xf32>, %matC: memref<1024x1024xf32>) workgroup(%sharedA : memref<32x32xf32, #gpu.address_space<workgroup>>, %sharedB : memref<32x32xf32, #gpu.address_space<workgroup>>) kernel {
```
linuxlonelyeagle, line 3: I think there needs to be some comments in this example, pure code doesn't feel very good. Like how you did matmul,some details of the code.

Xinyu302: I have added some comments. However, the matmul case is still not easy to understand. The top answer to this [Stackoverflow question](https://stackoverflow.com/questions/18815489/cuda-tiled-matrix-matrix-multiplication-with-shared-memory-and-matrix-size-whic) described what I do.

PR #159, opened 2023-06-07
[thirdparty] change std::int32_t to int32_t

Avimitin:
This fixed the compile error on latest compiler toolchain.


zhanghb97:
Thanks!

PR #160, opened 2023-06-07
[chore] add .cache directory into .gitignore

Avimitin:
The clangd lsp server will create a .cache directory in the buddy-mlir root directory which shouldn't be tracked by git.


zhanghb97:
Thanks!

PR #172, opened 2023-07-04
[examplese][MLIRSparseTensor] Add initial example for binary operation

Avimitin:
- [x] `func.func @intersection`
- [x] keyword `identity`
- [x] `func.func @union`


zhanghb97 (approved):
Thanks! Please resolve the conflict and push to the main branch!

PR #173, opened 2023-07-05
[MLIRPython Enhancement] Optimized Importer for Easier Operator Implementation in PyTorch Dynamo

xTayEx:
This PR improves the implementation of the importer in the buddy compiler for the PyTorch Dynamo backend. The optimization makes the implementation of operators with multiple results easier.

In this PR, the `symbol_table` is a dictionary with keys being tuples that consist of the operator name and the index of result of the operator. For example, `(var_mean, 0): op` means that the 0th result of `var_mean` operator is produced by `op`. In this way, we can decompose an operator with multiple results into multiple operators with single result. Still taking `var_mean` as an example. According to the documentation of PyTorch, this operator has two results: `var` and `mean`. We can decompose it into `varOp` and `meanOp`, and the two results is recorded as `(var_mean, 0): varOp` and `(var_mean, 1): meanOp` in `symbol_table`


zhanghb97 (commented):
Thanks for this patch! We are currently exploring the integration under the `examples` dir, and when the process becomes complete, we will integrate all of this part into the frontend path. Therefore, we need to standardize the code style from now on and improve the documentation as we develop.

- Let's add some docstrings! According to PEP8 and PEP257: "Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the def line."
- Please use a code format tool to reformat the code. I tend to use `black`, is there a more recommended tool?
- How about writing some description about the design into the `examples/MLIRPython/README.md`?

xTayEx:
> Thanks for this patch! We are currently exploring the integration under the `examples` dir, and when the process becomes complete, we will integrate all of this part into the frontend path. Therefore, we need to standardize the code style from now on and improve the documentation as we develop.
> 
>     * Let's add some docstrings! According to PEP8 and PEP257: "Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the def line."
> 
>     * Please use a code format tool to reformat the code. I tend to use `black`, is there a more recommended tool?
> 
>     * How about writing some description about the design into the `examples/MLIRPython/README.md`?

Thank you for your feedback! I am excited about the opportunity to contribute to buddy-compiler!
To address your points:
1. I agree with you completely. I will follow PEP8 and PEP257 to add some docstrings and make new commits
2. In fact, I have already formatted the code using `black`, which I regularly use as my code reformat tool. If there are specific sections of the code you believe require further formatting, please let me know. I'm more than happy to follow to the project's code style guidelines.
3. I agree with your suggestion. I will add some description about the importer design in README in new commit. btw, could you let me know in which part you'd like to see this in the README.md? I want to make sure it meshes well with everything else that's already there.

xTayEx:
Got it! Here are the updates:
1. Add docstrings in Google style for all public functions and classes.
2. Reformat the code with `yapf` and MLIR's custom style configuration.
3. Add methodology part in README.md to describe the design.
4. Add `addmm` operator for demo purpose.

Give the updates a review when you're free. Thanks!

zhanghb97:
Gooooood! Thanks!

PR #180, opened 2023-08-05
[examples][MLIRPython] support different tensor dtype in importer

xTayEx:
Currently, importer only support tensor with `torch.float32` dtype. This PR adds support for different tensor dtype.


examples/MLIRPython/buddy/compiler.py
```diff
@@ -72,15 +72,22 @@ def import_graph(self) -> ir.Module:
...
       for arg in self._inputs:
         shape_list = list(arg.shape)
-        f32 = ir.F32Type.get()
-        tensor_arg = ir.RankedTensorType.get(shape_list, f32)
+        dtype = arg.dtype
+        if dtype is torch.int32:
+          mlir_dtype = ir.IntegerType.get_signless(32)
+        elif dtype is torch.float32:
+          mlir_dtype = ir.F32Type.get()
+        else:
```
linuxlonelyeagle, line 87: I think it would be more elegant to use a statement like switch here, since more types may be supported later, and I think you can refer to the match statement.

linuxlonelyeagle (commented):
I made some brief comments. I think the corresponding test cases should also be provided .

xTayEx:
> I made some brief comments. I think the corresponding test cases should also be provided .

Got it! I have add some test and use pattern match syntax in the new update. Give a review when you're free. Thanks :)

linuxlonelyeagle:
When an issue is resolved, you can click `Resolve conversation`.

examples/MLIRPython/addmm.py
```diff
@@ -8,7 +8,15 @@ def foo(c, a, b):
...
+b_float32 = torch.randn(2, 3)
+c_float32 = torch.randn(3, 3)
+foo_mlir(c_float32, a_float32, b_float32)
+
+a_int32 = torch.randint(10, (3, 2)).to(torch.int32)
+b_int32 = torch.randint(10, (2, 3)).to(torch.int32)
+c_int32 = torch.randint(10, (3, 3)).to(torch.int32)
+foo_mlir(c_int32, a_int32, b_int32)
+
+
```
linuxlonelyeagle, line 22: Delete an endline

examples/MLIRPython/addmm.py
```diff
@@ -8,7 +8,15 @@ def foo(c, a, b):
```
linuxlonelyeagle, line 9: Is there a redundant endline here as well?

examples/MLIRPython/test/test_different_dtype.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+foo_mlir = dynamo.optimize(compiler.DynamoCompiler)(foo)
+float32_in1 = torch.randn(10).to(torch.float32)
+float32_in2 = torch.randn(10).to(torch.float32)
+foo_mlir(float32_in1, float32_in2)
+
+int32_in1 = torch.randint(0, 10, (10,)).to(torch.int32)
+int32_in2 = torch.randint(0, 10, (10,)).to(torch.int32)
+foo_mlir(int32_in1, int32_in2)
```
linuxlonelyeagle, line 17: You should not create a new test directory here, you should write the code here to arith_add.py.

xTayEx: Thank you for you review. I think the `arith_add.py` file is for demo purpose, not for testing. Therefore I create the `test_different_dtype.py` for just for testing the different dtype support feature. What do you think?

linuxlonelyeagle: I see what you mean, but I don't think it makes sense to recreate a directory for testing purposes, and at the same time it would cause confusion in the directory structure. `arith_add.py` should be used as an example and also for testing purposes.I think this is an obvious thing to do.

examples/MLIRPython/arith_add.py
```diff
@@ -2,10 +2,16 @@
 import torch
 import torch._dynamo as dynamo
 
+
 def foo(x, y):
   return x + y
 
+
```
linuxlonelyeagle, line 9: The two new lines added in rows 5 and 9 look unnecessary.

linuxlonelyeagle (approved):
Looks good, there are a few minor problems.

examples/MLIRPython/arith_add.py
linuxlonelyeagle: OK, it doesn't look like it needs to be changed.

zhanghb97:
Thanks! @xTayEx @linuxlonelyeagle

PR #184, opened 2023-08-15
[DIP] Change resize2D interface

FlagerLee:
Change resize2D (in DIP.h) interface from
`inline MemRef<float, 2> Resize2D(Img<float, 2> *input, INTERPOLATION_TYPE type, std::vector<float> scalingRatios)`
to
`inline MemRef<float, 2> Resize2D(Img<float, 2> *input, INTERPOLATION_TYPE type, std::vector<uint> size)`
Now resize2D passes new image size instead of scaling ratio.


zhanghb97 (approved):
Thanks!

zhanghb97:
ping @FlagerLee

meshtag:
Our design was already capable of working with target image sizes directly, scaling ratios were not compulsorily required. 

OpenCV's python resize API can also optionally accept scaling ratios.

meshtag:
I fear this change makes the interface lose some of its capabilities and would make our API diverge from the OpenCV one. If this change is not justified, I suggest reverting it. 

If you are unhappy about the intermediate floating point calculations, we can definitely find an alternative route for that. But I am not in favour of removing the option to provide scaling ratios.

FlagerLee:
I see. I missed the function accept image size. But another question is these two functions' difference is the last argument type, which I don't think it is a good design(but of course this is another question).
You are right, please revert this commit @zhanghb97. Sorry for this.

FlagerLee:
Besides, I recommend change Resize2D return type from Memref to Img. Do you have any thoughts on this?

FlagerLee:
> I think we can change the return type for C++ interface of `Resize2D` to `img` from `memref`.

Yes, that's exactly what I want to do.

PR #187, opened 2023-08-19
[midend][tests] Add broadcast BatchMatMul optimization pass and tests.

EllisLambda:
Add "-batchmatmul-optimize" option in buddy-opt, the Matmul part is similar to [MatMulBroadcast.mlir](https://github.com/buddy-compiler/buddy-benchmark/blob/main/benchmarks/OpOptimization/MatMul/MatMulBroadcast.mlir) in buddy-benchmark. The batch-level loop utilizes affine.parallel. In the case of multiple batches, OpenMP can be utilized for multi-threaded acceleration. The A[batch_idx, 0, 0] which broadcasts in the Matmul will use `affine.prefetch` to prefetch in the outer loop. It could make the Op slightly faster. There's a bug, `if (A_elementType.isIntOrFloat())` cannot work properly. Even if the input type is floating-point, the integer branch will still execute. So I modified to `if (A_elementType.isIntOrFloat() && 0)` to skip the branch. So far the integer matrix is unavailable.

### DIFF
--- a/examples/MLIRLinalg/linalg-batch-matmul.mlir
+++ b/examples/MLIRLinalg/linalg-batch-matmul.mlir
@@ -1,0 +1,29 @@
+// RUN: buddy-opt -batchmatmul-optimize -verify-diagnostics -expand-strided-metadata -lower-affine -convert-vector-to-llvm -finalize-memref-to-llvm -convert-scf-to-cf -convert-linalg-to-llvm -llvm-request-c-wrappers -convert-func-to-llvm -reconcile-unrealized-casts %s \
+// RUN: | mlir-cpu-runner -O0 -e buddy_batchmatmul_f32 \
+// RUN: -shared-libs=%mlir_runner_utils_dir/libmlir_runner_utils%shlibext,%mlir_runner_utils_dir/libmlir_c_runner_utils%shlibext \
+// RUN: | FileCheck %s
+
+memref.global "private" @A : memref<2x2x3xf32> = dense<[[[9., 4., 6.],[2., 4., 0.]],[[6., 3., 3.],[0., 4., 7.]]]>
+memref.global "private" @B : memref<2x3x4xf32> = dense<[[[1., 3., 8., 0.],[1., 8., 8., 7.], [6., 9., 7., 9.]],[[3., 8., 6., 8.],[2., 7., 0., 6.],[0., 4., 0., 4.]]]>
+memref.global "private" @C : memref<2x2x4xf32> = dense<[[[ 49., 113., 146.,  82.],[  6.,  38.,  48.,  28.]],[[ 24.,  81.,  36.,  78.],[  8.,  56.,   0.,  52.]]]>
+
+func.func private @printMemrefF32(memref<*xf32>) attributes { llvm.emit_c_interface }
+
+func.func @buddy_batchmatmul_f32() -> f32{
+  %a = memref.get_global @A : memref<2x2x3xf32>
+  %b = memref.get_global @B : memref<2x3x4xf32>
+  %c = memref.get_global @C : memref<2x2x4xf32>
+
+  linalg.batch_matmul 
+      ins(%a, %b: memref<2x2x3xf32>, memref<2x3x4xf32>)
+      outs(%c: memref<2x2x4xf32>)
+  %printed_c = memref.cast %c : memref<2x2x4xf32> to memref<*xf32>
+  call @printMemrefF32(%printed_c) : (memref<*xf32>) -> ()
+  // CHECK: {{Unranked Memref base@ = 0x[0-9A-Fa-f]{1,} rank = 3 offset = 0 sizes = \[2, 2, 4\] strides = \[8, 4, 1\] data =}}
+  // CHECK{LITERAL}: [[[98,    226,    292,    164], 
+  // CHECK{LITERAL}:   [12,    76,    96,    56]], 
+  // CHECK{LITERAL}:  [[48,    162,    72,    156], 
+  // CHECK{LITERAL}:   [16,    112,    0,    104]]]
+  %zero = arith.constant 0.0 :f32
+  return %zero :f32
+}
--- a/examples/MLIRLinalg/makefile
+++ b/examples/MLIRLinalg/makefile
@@ -136,6 +136,61 @@ linalg-matmul-optimize-run:
 	${MLIR_CPU_RUNNER} ${OPT_FLAG} -e main -entry-point-result=void \
 		-shared-libs=${MLIR_RUNNER_UTILS} -shared-libs=${MLIR_C_RUNNER_UTILS}
 
+linalg-batch-matmul-optimize-run:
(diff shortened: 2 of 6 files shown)


midend/lib/Conversion/MatMulOptimization/BatchMatMulOptimize.cpp
```diff
@@ -0,0 +1,282 @@
+//===- BatchMatMulOptimize.cpp
+//-------------------------------------------------===//
```
xlinsist, line 2: Please reformat the title comment for neatness.

xlinsist (changes requested):

midend/lib/Conversion/MatMulOptimization/BatchMatMulOptimize.cpp
```diff
@@ -0,0 +1,282 @@
...
+    const Value c0_dynamicType_vec = rewriter.create<vector::SplatOp>(
+        loc, VectorType::get({stepPlaceHolder}, A_elementType), c0_dynamicType);
+
+    // Dims
+    Value BATCH = rewriter.create<memref::DimOp>(loc, A, 0); // batch size
+    Value M = rewriter.create<memref::DimOp>(loc, A, 1);     // a_row
+    Value N = rewriter.create<memref::DimOp>(loc, B, 2);     // b_col
+    Value K = rewriter.create<memref::DimOp>(loc, B, 1);     // b_row
+
+    // build loop body
```
xlinsist, line 81: Please make sure the comments are decent(i.e. capitalized and expressed formally). You should also double-check the other comments in this PR.

examples/MLIRLinalg/linalg-batch-matmul.mlir
```diff
@@ -0,0 +1,29 @@
...
+  %printed_c = memref.cast %c : memref<2x2x4xf32> to memref<*xf32>
+  call @printMemrefF32(%printed_c) : (memref<*xf32>) -> ()
+  // CHECK: {{Unranked Memref base@ = 0x[0-9A-Fa-f]{1,} rank = 3 offset = 0 sizes = \[2, 2, 4\] strides = \[8, 4, 1\] data =}}
+  // CHECK{LITERAL}: [[[98,    226,    292,    164], 
+  // CHECK{LITERAL}:   [12,    76,    96,    56]], 
+  // CHECK{LITERAL}:  [[48,    162,    72,    156], 
+  // CHECK{LITERAL}:   [16,    112,    0,    104]]]
+  %zero = arith.constant 0.0 :f32
+  return %zero :f32
+}
```
xlinsist, line 29: Add a space at the end of the line.

examples/MLIRLinalg/makefile
```diff
@@ -136,6 +136,76 @@ linalg-matmul-optimize-run:
...
+		-lower-affine \
+		-convert-scf-to-cf \
+		-convert-vector-to-llvm \
+		-finalize-memref-to-llvm \
+		-convert-arith-to-llvm \
+		-convert-func-to-llvm \
+		-reconcile-unrealized-casts | \
+	${MLIR_TRANSLATE} --mlir-to-llvmir -o log.ll
+
+linalg-batch-matmul-optimize-run:
```
xlinsist, line 194: In line 194, `linalg-batch-matmul-optimize-run` is redundant with line 137 and should be removed.

examples/MLIRLinalg/makefile
```diff
@@ -136,6 +136,76 @@ linalg-matmul-optimize-run:
 	${MLIR_CPU_RUNNER} ${OPT_FLAG} -e main -entry-point-result=void \
 		-shared-libs=${MLIR_RUNNER_UTILS} -shared-libs=${MLIR_C_RUNNER_UTILS}
 
+linalg-batch-matmul-optimize-run:
+	@${BUDDY_OPT} linalg-matmul.mlir ${MLIR_OPT_OPTIONS} \
+		-batchmatmul-optimize="step-placeholder=64" \
```
xlinsist, line 141: It is recommanded to use `vector-size` instead of  `step-placeholder` since it is easyer to understand.

xlinsist (changes requested):

xlinsist:
Thanks! @EllisLambda I have left some suggestions to make it better, and I think these are enough for this PR. More verifications in real scenarios(e.g. simultaneously process `linalg.batch_matmul` of multiple sizes) may be needed in the future.

@zhanghb97 you may also see https://github.com/buddy-compiler/buddy-benchmark/pull/73.

xlinsist:
> @EllisLambda Why add batch_matmul pass to opt? I think it is inappropriate to add a certain op pass in opt.

`batch_matmul` is a typical opertion which is widely used in transformer models, so I think it is reasonable to provide a tailor-made pass for it.

Nevertheless, it would be better if `matmul-optimize` pass can incorporate the lowering optimizations of both `linalg.matmul` and `linalg.batch_matmul`. WDYT @EllisLambda ?

linuxlonelyeagle:
In fact, there is some truth to this, but I think this PR should explain what optimizations have been made? Where has the optimization been made?

EllisLambda:
> In fact, there is some truth to this, but I think this PR should explain what optimizations have been made? Where has the optimization been made?

I tried to 

> WDYT

The optimization styles between Matmul and BatchMatmul are quite distinct. On different hardware, the performance of these two operations may differ. Separating might be better?

EllisLambda:
> In fact, there is some truth to this, but I think this PR should explain what optimizations have been made? Where has the optimization been made?



> > @EllisLambda Why add batch_matmul pass to opt? I think it is inappropriate to add a certain op pass in opt.
> 
> `batch_matmul` is a typical opertion which is widely used in transformer models, so I think it is reasonable to provide a tailor-made pass for it.
> 
> Nevertheless, it would be better if `matmul-optimize` pass can incorporate the lowering optimizations of both `linalg.matmul` and `linalg.batch_matmul`. WDYT @EllisLambda ?

The optimization styles between Matmul and BatchMatmul are quite distinct. On different hardware, the performance of these two operations may differ. Separating might be better?

EllisLambda:
> In fact, there is some truth to this, but I think this PR should explain what optimizations have been made? Where has the optimization been made?

Ok, I will add the explanation.

xlinsist:
> Separating might be better? 
Yes, we decide to develop `batchmatmul-optimize` pass specifically for `linalg.batch_matmul` right now.

Yumin@EllisLambda , about the "bug" you mention in your description, we can unify the support of floating point and integer with another interface. If not, starting a new discussion about integer operations would be more appropriate so that the relevant comments in your description could be removed, as this PR just needs to take floating point operations into account.

EllisLambda:
> > Separating might be better?
> 
> Yes, we decide to develop `batchmatmul-optimize` pass specifically for `linalg.batch_matmul` right now.
> 
> Yumin @EllisLambda , about the "bug" you mention in your description, we can unify the support of floating point and integer with another interface. If not, starting a new discussion about integer operations would be more appropriate so that the relevant comments in your description could be removed, as this PR just needs to take floating point operations into account.

Ok! I see.

EllisLambda:
@xlinsist I have amended my PR.

PR #188, opened 2023-08-21
[DIP] Reconstruct resize2D.

FlagerLee:
Reconstruct resize2D operation. Current op acts the same as opencv.


meshtag (changes requested):
Please resolve the merge conflict and update this PR with current version of buddy-mlir.

meshtag:
Please make the API compatible with current Img container in buddy-mlir. The PR can't be benchmarked in its current state and we can't determine the performance improvement due to it.

PR #193, opened 2023-09-01
[cmake] fix llvm out-of-tree build for one-step build strategy

Avimitin:
The current one-step build strategy requires LLVM sources to be placed in the buddy-mlir project and named 'llvm'. This makes packaging a difficult task, as users can't build buddy-mlir with other working llvm source code that has a different version to the one provided by buddy-mlir.

This commit sets 'LLVM_PROJECT_SOURCE_DIR' to be the up dir of 'LLVM_MAIN_SRC_DIR'. 'LLVM_MAIN_SRC_DIR' is set when building within the llvm source directory, so it is safe to use it as the relative path to llvm project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,11 +62,9 @@ else()
     # MLIR/LLVM Configuration
     #-------------------------------------------------------------------------------
 
-    # Allow passing external LLVM source, instead of forcing user using the vendored one
-    if (NOT LLVM_PROJECT_SOURCE_DIR)
-      set(LLVM_PROJECT_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/llvm")
-      message(STATUS "Using LLVM Project ${LLVM_PROJECT_SOURCE_DIR}")
-    endif()
+    # Allow using out-of-tree llvm directory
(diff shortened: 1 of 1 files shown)


linuxlonelyeagle:
To be honest, I think this would lead to the problem that using a different version of `LLVM` wouldn't compile the code, and in this case, every update of `LLVM` would require us to modify the `buddy-mlir` code to adapt it to the new version of `LLVM`.

Avimitin:
No, it won't. The buddy-mlir project vendored llvm is already an announcement to other users that this is the only supported version of the llvm project. So if someone wants to use a different llvm source, it's their responsibility to keep llvm and buddy-mlir in sync. For me, it will be much easiler to package buddy-mlir in some traditional package managers. For normal users who just want to use budd-mlir to test and build something, this PR doesn't affect the normal git submodule workflow. Instead, the current build system doesn't need to find out where the llvm source code is: we already run those build commands inside the llvm worktree.

Avimitin:
@zhanghb97 how do you think about this PR? It doesn't affect our current workflow but can also benefit other packagers.

zhanghb97:
@Avimitin Thanks! Sorry for the late reply!

PR #200, opened 2023-09-11
[build system]fix build bug.

linuxlonelyeagle:
* Fixed a bug in the build process.
* Changed some formatting.

### DIFF
--- a/midend/include/Dialect/Gemmini/Transform.h
+++ b/midend/include/Dialect/Gemmini/Transform.h
@@ -50,7 +50,7 @@ void populateGemminiLegalizeForLLVMExportPatterns(LLVMTypeConverter &converter,
                                                   RewritePatternSet &patterns,
                                                   int64_t dim, int64_t addrLen, 
                                                   size_t sizeOfElemT, size_t sizeOfAccT);
-void configureGemminiegalizeForExportTarget(LLVMConversionTarget &target);
+void configureGemminiLegalizeForExportTarget(LLVMConversionTarget &target);
 
 } // namespace mlir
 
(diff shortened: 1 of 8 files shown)


FlagerLee:
ContainerTest.cpp L111 should also be changed?
Why removing BatchMatMulOptimization?

xlinsist:
FYI: A latest landed PR about BatchMatMulOptimization pass https://github.com/buddy-compiler/buddy-mlir/pull/187 has problems in cmake configuration and should be solved by https://github.com/buddy-compiler/buddy-mlir/pull/199.

cc @EllisLambda

FlagerLee:
1. BatchMatMul bug solved by #199 
2. The role of 'overloading bracket operator test' in ContainerTest is unclear. I'll fix it after discussed with @zhanghb97 (probably by deleting this test).

linuxlonelyeagle:
Is it necessary to compile and generate two libs? I just asked him to compile it into a lib, which is consistent with the directory name where these two files are stored.
`BatchMatMulOptimize.cpp` is compiled together with `MatMulOptimize.cpp` to generate the `MatMulOptimization`. This doesn't look very elegant, but I'm just making some suggestions.
```
add_mlir_library(MatMulOptimization
  BatchMatMulOptimize.cpp
	MatMulOptimize.cpp
)
```

FlagerLee:
@xlinsist

xlinsist:
> Is it necessary to compile and generate two libs? I just asked him to compile it into a lib, which is consistent with the directory name where these two files are stored. `BatchMatMulOptimize.cpp` is compiled together with `MatMulOptimize.cpp` to generate the `MatMulOptimization`. This doesn't look very elegant, but I'm just making some suggestions.
> 
> ```
> add_mlir_library(MatMulOptimization
>   BatchMatMulOptimize.cpp
> 	MatMulOptimize.cpp
> )
> ```

Thanks for your suggestions! We are considering combining two optimizations into one pass since batch matmul is one special case of matmul, so this part may be subject to change in the near future.

PR #201, opened 2023-09-12
[midend][examples] Correct parameter and variable `stepPlaceholder` in batchmatmul optimization.

EllisLambda:


examples/MLIRLinalg/makefile
```diff
@@ -137,8 +137,12 @@ linalg-matmul-optimize-run:
 		-shared-libs=${MLIR_RUNNER_UTILS} -shared-libs=${MLIR_C_RUNNER_UTILS}
 
 linalg-batch-matmul-optimize-run:
+<<<<<<< HEAD
```
xlinsist, line 140: Remove these traces.

examples/MLIRLinalg/makefile
```diff
@@ -152,34 +156,97 @@ linalg-batch-matmul-optimize-run:
 		-shared-libs=${MLIR_RUNNER_UTILS} -shared-libs=${MLIR_C_RUNNER_UTILS}
 
 linalg-batch-matmul-lower:
-	@${MLIR_OPT} linalg-batch-matmul.mlir ${MLIR_OPT_OPTIONS} \
+	@${MLIR_OPT} linalg-batch-matmul_f32.mlir ${MLIR_OPT_OPTIONS} \
```
xlinsist, line 159: Use `linalg-batch-matmul-f32` instead of `linalg-batch-matmul_f32` and modify the file name similarly.

xlinsist:
Quick fix! The last question: It seems that we don't need to add additional test cases for i8. It looks cumbersome in the case that the algorithm and lowering step are the same as when dealing with f32. Under normal circumstances, we expect that this pass can deal with both so no test examples is needed to inform users that we can support both.

EllisLambda:
@xlinsist Test cases might make modifying buddy-opt more convenient, is keep the i8 mlir file and delete the tests in makefile a good idea?

xlinsist:
> @xlinsist Test cases might make modifying buddy-opt more convenient, is keep the i8 mlir file and delete the tests in makefile a good idea?

That's alright, it is not recommanded to keep files that are not used or mentioned in the project, so we can keep the i8 test. But I think the parameter name `step` should be modfied.

EllisLambda:
@xlinsist The parameter name `step` has been changed.

xlinsist (approved):

midend/lib/Conversion/MatMulOptimization/BatchMatMulOptimize.cpp
```diff
@@ -279,9 +279,9 @@ class BatchMatMulOptimizePass
                     affine::AffineDialect, VectorDialect>();
   }
 
-  Option<int64_t> stepPlaceHolder{
-      *this, "step-placeholder",
-      llvm::cl::desc("Affine step placeholder size."), llvm::cl::init(64)};
+  Option<int64_t> affineStep{
+      *this, "step",
+      llvm::cl::desc("Affine step size."), llvm::cl::init(64)};
```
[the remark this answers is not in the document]
EllisLambda: @xlinsist Fixed.

xlinsist:
LGTM, thanks!

PR #205, opened 2023-09-21
[DIP][OSPP] Add JPEG image encoding and decoding.

Guan-schoolmate:


FlagerLee (commented):
I'll check its correctness later.

FlagerLee:
Correctness proved

frontend/Interfaces/buddy/DIP/imgcodecs/grfmt_jpeg.h
```diff
@@ -0,0 +1,764 @@
+//===- Grfmt_jpeg.h ---------------------------------------------------===//
```
zhanghb97, line 1: Wrong format:
- Grfmt_jpeg.h -> grfmt_jpeg.h
- 80-column format.

zhanghb97 (commented):
Do you think we should add some JPEG-specific examples and tests in this PR?

FlagerLee:
You need to add new test file since we need bmp, jpeg and png tests, not just one of them.

examples/DIPDialect/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 if(BUDDY_ENABLE_OPENCV)
   find_package(OpenCV REQUIRED CONFIG)
```
FlagerLee, line 2: This condition is BUDDY_ENABLE_OPENCV, not enable jpeg or etc. . Can we bind libjpeg, libpng, libbmp and BuddyLibDIP together so that we only need to link one library(or variable such as ${DIP_LIBS})?

FlagerLee (changes requested):

FlagerLee (approved):

PR #208, opened 2023-09-21
[OSPP] Implementation of TorchDynamo compiler

xTayEx:


frontend/Python/ops/tosa.py
```diff
@@ -118,12 +118,71 @@ def _normalize_binary_operator_args(arg1, arg2):
         )
 
 
+def addmm_op(
+    node, symbol_table: Dict[Tuple[str, int], ir.Operation]
+) -> ir.Operation:
```
zhanghb97, line 123: Please add docstrings here. I think it should consist of the following parts:
- **Description**: describe the conversion from PyTorch operator to MLIR operation.
- **Note**: If necessary, list some notes during the conversion process.
- **Args and Returns**: If they are all repetitive, we can only write for the first conversion function, and the remaining functions can omit these two items.

For example:

```
"""
  Import matrix multiplication operation.
  From PyTorch `torch.addmm` operator to MLIR TOSA `matmul` operation.
  
  Note: this function first reshapes the input matrices to 3D tensors
  (since tosa.MatMulOp requires it). Then it multiplies these reshaped matrices.
  Finally, it adds the input tensor to the matrix multiplication result.

  Args:
  - node: Containing information from the input graph node.
  - symbol_table: A dictionary mapping symbols to their corresponding operations.

  Returns:
  - op: The operation representing the result of adding the matrix multiplication 
        to the input tensor.
  """
```

zhanghb97 (commented):
@xTayEx Nice patch!
We also need to add the following items:
- Docstrings for conversion functions (see the example for `addmm`)
- Tests for these conversion functions (see [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/tests/Python/test_arith_add.py) as an example)
- Add a README for BuddyPython and describe the design points and import strategy.

xTayEx:
@zhanghb97 Thank you for your reply! I have updated just now and the format issues should be fixed

zhanghb97:
Congrats 🎉🎉🎉
Thank you for your contribution. Hope you enjoyed the OSPP project!

PR #216, opened 2023-09-29
[OSPP23] Add llama2 lowering end2end

weilinquan8:


weilinquan:
test script and readme.md are in examples/MLIRLlama/

zhanghb97 (commented):
@weilinquan Nice!

We also need to add the following items:

- Docstrings for conversion functions (see the example for [addmm](https://github.com/buddy-compiler/buddy-mlir/pull/208))
- Tests for these conversion functions (see [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/tests/Python/test_arith_add.py) as an example)
- The current example process is somewhat complex, and we expect to use CMake to integrate the various processes, which we will discuss in detail at our group meeting.

examples/BuddyLlama/llama-main.cpp
```diff
@@ -36,45 +37,44 @@ int main() {
...
-  clock_t buddyTokenizeEnd = clock();
-  double buddyTokenizeTime =
-      (double)(buddyTokenizeEnd - buddyTokenizeStart) / CLOCKS_PER_SEC;
+  auto buddyTokenizeEnd = system_clock::now();
+  auto buddyTokenizeTime = duration_cast<seconds>(buddyTokenizeEnd - buddyTokenizeStart);
   // Print the tokenized result
   cout << "Get User input:" << pureStrContainer.revert(pureStrContainer)
        << endl;
-  cout << "[Buddy] Tokenize input time: " << buddyTokenizeTime * 1000 << "ms"
+  cout << "[Buddy] Tokenize input time: " << buddyTokenizeTime.count() * 1000 << "ms"
```
Lester-1, line 48: What is the unit of time measurement here?
The unit multiplied by 1000 here is “ms”, and the unit divided by 1000 below is “s”.

weilinquan: Yes! I have change buddyTokenizeTime's time unit to milliseconds and update this cout.

examples/BuddyLlama/test-llama2.py
```diff
@@ -0,0 +1,28 @@
...
+import torch._dynamo as dynamo
+from torch._inductor.decomposition import decompositions as inductor_decomp
+from torch._functorch.aot_autograd import aot_autograd_decompositions
+import numpy
+import os
+
+from buddy.compiler.frontend import DynamoCompiler
+from buddy.compiler.ops import tosa
+
+tokenizer = LlamaTokenizer.from_pretrained('/llama-2-7B-hf')
```
zhanghb97, line 12: The `/llama-2-7B-hf` seems to cause an error: 
```
huggingface_hub.utils._validators.HFValidationError: 
Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are forbidden, 
'-' and '.' cannot start or end the name, max length is 96: '/llama-2-7B-hf'.
```
Did I miss something important, or should we remove the `/`?

zhanghb97: Ah, I recalled the details about the configurations here. We should modify the path to the huggingface version of the llama model, right?

weilinquan: Yes

frontend/Python/frontend.py
```diff
@@ -241,7 +275,7 @@ def generated_func(*args):
                         for output_arg in output_node_args:
                             op = self._symbol_table.get((str(output_arg), 0))
                             returns.append(op)
-
+                        returns = returns[0]
```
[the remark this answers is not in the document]
weilinquan: In llama2，it will return many tensors include intermediate tensors for compute gradient, but we only need the first tensor to generate next token.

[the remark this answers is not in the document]
weilinquan: Maybe we should use a better way to solve it

examples/BuddyLlama/test-llama2.py
weilinquan: I will change it from '/llama-2-7B-hf' to 'path to huggingface llama2 model'

zhanghb97 (approved):
LGTM! Congrats 🎉🎉🎉
Thank you for your contribution. Hope you enjoyed the OSPP project!

PR #225, opened 2023-10-29
[DAP] Add iir_simd operation for DAP dialect.

taiqzheng:
This is the **SIMD version** for **iir operation** in DAP dialect. 

Currently, only support examples with `num{SOS matrix} <= 16`. Also, this `dap.iir_simd` only support data with type f64(double). Please check this PR with the new iir benchmarks(BuddyIirBenchmark.cpp & KFRIir.cpp [PR in buddy-benchmark](https://github.com/buddy-compiler/buddy-benchmark/pull/111)

So, future work:
1.  Add attribute to support `num{SOS matrix} > 16`
2. Extract common program section to Utils(maybe a DAPUtils.cpp is needed)


taiqzheng:
Buddy benchmark PR#110 changed to PR#111.

PR #227, opened 2023-10-31
fix softmax error

weilinquan8:


examples/BuddyLlama/CMakeLists.txt
```diff
@@ -63,14 +63,4 @@ SET_TARGET_PROPERTIES(
...
-  LLAMA
-  mlir_c_runner_utils
-  omp
-)
-if(BUDDY_MLIR_USE_MIMALLOC)
-  list(APPEND BUDDY_LLAMA_LIBS mimalloc)
-endif()
-
-target_link_libraries(buddy-llama-run ${BUDDY_LLAMA_LIBS})
+target_link_libraries(buddy-llama-run LLAMA)
```
zhanghb97, line 66: The `LLAMA` target linked libraries retain the previous version. (use `BUDDY_LLAMA_LIBS`)

examples/BuddyLlama/llama-main.cpp
```diff
@@ -20,13 +20,15 @@
...
+#include <limits>
+#include <type_traits>
 
 using namespace buddy;
 using namespace std;
 using namespace chrono;
 
 extern "C" void _mlir_ciface_forward(MemRef<float, 3> *, MemRef<float, 1> *,
-                                     MemRef<size_t, 2> *);
+                                     MemRef<long long, 2> *);
```
zhanghb97, line 31: I think `size_t` is the more suitable option here. What is the advantage of using `long long`?

weilinquan8: Yes, I will update this.

examples/BuddyLlama/CMakeLists.txt
weilinquan8: OK

tests/Python/test_t.py
```diff
@@ -25,8 +25,13 @@ def foo(x):
 
 # CHECK: module {
 # CHECK-LABEL: func.func @forward
+<<<<<<< HEAD
```
zhanghb97, line 28: Wrong code. Please remove these. Same thing in the following file.

examples/BuddyLlama/CMakeLists.txt
```diff
@@ -48,29 +47,29 @@ add_custom_command(
...
-  mlir_c_runner_utils
-  omp
-)
-if(BUDDY_MLIR_USE_MIMALLOC)
-  list(APPEND BUDDY_LLAMA_LIBS mimalloc)
-endif()
+  add_library(LLAMA STATIC llama.o)
 
-target_link_libraries(buddy-llama-run ${BUDDY_LLAMA_LIBS})
+  SET_SOURCE_FILES_PROPERTIES(
```
zhanghb97, line 52: It seems we can remove this `SET_SOURCE_FILES_PROPERTIES`.

zhanghb97 (commented):
I recommend that we split the container modification into another PR.

PR #231, opened 2023-11-04
[frontend] Replace `is_inference` with `torch.no_grad` for more elegant FX Graph output node arguments handling

xTayEx:


zhanghb97 (commented):
Please modify the BuddyLlama example with `torch.no_grad` as well.

zhanghb97 (approved):
Thanks!

PR #247, opened 2023-11-28
[frontend[ fix Python install

Avimitin:
This PR use the LLVM MLIR Python CMake macro to build the Python package. This can help us get the Python interface after cmake install.


zhanghb97:
Reusing MLIR's cmake interface is a great idea! However, let's place Buddy's Python package in our own build path (try to give the `ROOT_DIR`). In the future, this will also be the location for our frontend, midend, and backend Python interfaces. Putting our Python package in the LLVM build path could confuse users and prove more challenging to maintain.

Avimitin:
Actually it is in our own build path, the updated document is using one-step build strategy, so it seems like our python packages is store inside llvm build dir. I will update the document about two-step build strategy.

Avimitin:
Documents are updated to follow the two-step build strategy. Please review again.

frontend/Python/buddy/compiler/frontend.py
zhanghb97, line 1: Is this position changing to adapt to a one-step build? Can we keep the original position for these Python files?
`buddy-mlir/frontend/Python/buddy/compiler/frontend.py` seems too cumbersome.

Avimitin: No, actually this change is to make the directory structure consistent with python import path. What if we move the Python binding to top level?

Avimitin:
Sorry for not updating for so long, finally free from the massive backend jobs. After reviewing these changes, I still feel that using cmake to populate Python module is not a good idea. CMake itself is not a great programmable build system, populating this directory, add install target means a lot of hack. I tried to implement some, found that it is unmaintainable. Second is that splitting python modules into different directories and grouping them together is not easy to read and find code, also not great for development: user needs to run cmake generate first.

One way to do things right, even without making a long path, is to group all Python code in one top-level directory, like upstream LLVM MLIR does. Looking forward to your thoughts and ideas for improvement

PR #255, opened 2023-12-06
[frontend]: add install target for frontend headers

Avimitin:
This can benefit downstream projects by allowing them to easily include headers from a distributed installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,3 +209,15 @@ add_subdirectory(tests)
 add_custom_target(check-buddy
   DEPENDS check-examples check-tests
 )
+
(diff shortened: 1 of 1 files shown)


meshtag (commented):
Why not put this in the [parent directory CMakeLists.txt](https://github.com/buddy-compiler/buddy-mlir/blob/main/CMakeLists.txt)? 

That feels like a good place for project-wide operations, like building everything and then installing stuff globally.

Avimitin:
Ok I will fix it up and amend to the previous PR.

meshtag (approved):

zhanghb97:
Thanks! @Avimitin @meshtag

PR #257, opened 2023-12-07
[examples] Fix stand-alone conv-opt edge detection build

meshtag:
I was getting an error saying that cmake was not able to find the target `buddy-image-container-test` when I was trying to have a stand-alone build of the `edge-detection` example for testing `conv-opt`. 

This patch should fix that problem and users should be able to build `edge-detection` target without building any DIP target first.


zhanghb97 (commented):
Thanks!

PR #258, opened 2023-12-07
[Examples] Fix cmake build file python command

harrisonGPU:
When I use CMake to build the LLAMA example on my MacBook, it encounters a 'python not found' issue. However, when I use the command python import-llama2.py in terminal, it works correctly. So, I want to use ${Python3_EXECUTABLE} to replace 'python'，it works correctly.Please review!


TSWorld1314:
@zhanghb97 hi,hongbin,can you see this issue? Please review

PR #259, opened 2023-12-14
[Git] Add shallow option to speed up pulling code

TSWorld1314:
When I use the `git submodule update --init` command to clone a submodule, it is very slow! (Maybe I need a proxy!) However, I think we do not need the entire LLVM history or that of any other third-party submodule. Using shallow = true in the .gitmodules file for a submodule configuration will reduce data transfer: a shallow clone fetches only the most recent commits instead of the entire commit history of the repository, resulting in faster cloning operations.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,6 +2,8 @@
 	path = llvm
 	url = https://github.com/llvm/llvm-project.git
 	branch = main
+	shallow = true
 [submodule "thirdparty/mimalloc"]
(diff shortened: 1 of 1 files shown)


TSWorld1314:
hi @zhanghb97 ,this is my suggestion: I believe this patch will help contributors pull code more quickly!

zhanghb97:
Thanks!

PR #263, opened 2023-12-18
MLIRGPU Dialect examples

gxsoar:
1. I add gpu.alloc, gpu.dealloc, gpu.memcpy, gpu.memset, gpu.spmm, gpu-index-ops, gpu.shuffle, gpu.printf operations example. 
2. In the mlir document, the `gpu-to-cubin` pass has been deleted. I fix this bug and change the makefile


zhanghb97 (commented):
- Do not simply copy and paste the test from upstream. Modify it specifically according to the meaning you wish to convey through the example, then add comments and analysis. See [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/MLIRVector/vector-gather.mlir#L69) as an example.
- Add tests for each use case. See [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/MLIRVector/vector-fma.mlir#L1) as an example.

gxsoar:
> * Do not simply copy and paste the test from upstream. Modify it specifically according to the meaning you wish to convey through the example, then add comments and analysis. See [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/MLIRVector/vector-gather.mlir#L69) as an example.
> * Add tests for each use case. See [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/MLIRVector/vector-fma.mlir#L1) as an example.

Ok, I will rewrite some examples and add comments and analysis on them

gxsoar:
Ok, I will rewrite some examples and add some comments, analysis on them. @zhanghb97

PR #264, opened 2023-12-18
MLIROpenacc examples

gxsoar:
I copy some mlir openacc test to buddy-mlir. I can lowering acc dialect to llvm, but cannot translate and run them


zhanghb97 (commented):
- Do not simply copy and paste the test from upstream. Modify it specifically according to the meaning you wish to convey through the example, then add comments and analysis. See [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/MLIRVector/vector-gather.mlir#L69) as an example.
- Add tests for each use case. See [here](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/MLIRVector/vector-fma.mlir#L1) as an example.

PR #271, opened 2024-01-01
[Examples][TOSA] Improve Tosa Dialect examples

meshtag:
This PR intends to fix TOSA dialect examples by removing a no longer supported pass (`convert-linalg-to-llvm`) from the lowering pipelines. It also adds a simple target to get `linalg.generic` loops lowered from `tosa.resize`.


zhanghb97:
Thanks!

Issue #265, opened 2023-12-21
The `torch` version used in `requirements.txt` seems to be removed


BHbean:
**Describe the bug**
I am trying to reproduce the example under `examples/BuddyLlama`, but I found that the `torch` version used in `requirements.txt` seems to be removed.

**To Reproduce**
- Create a new conda env named `llama` and activate it
```bash
$ conda create -n llama
$ conda activate llama
```
-  Intall the packages needed
```bash
$ cd buddy-mlir
$ pip install -r requirements.txt
```
Then we can see the error.

**Expected behavior**
`torch` is successfully installed.

**Screenshots**
![image](https://github.com/buddy-compiler/buddy-mlir/assets/66934674/cc981d17-671c-497b-86f7-78c9c56f53ce)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 22.04

**Additional context**
I am trying another "dev" version of `torch`. However, for the sake of generality, maybe we should use a stable version of `torch` to make the reproduction of this example more stable?

BHbean:
Besides, it seems that the `pybind11` package needs to be installed extraly by running `conda install pybind11 -c conda-forge` before building `llvm` and `buddy-mlir`, which will not be installed by running `pip install -r requirements.txt`.

zhanghb97:
ping @Lester-1

Lester-1:
Thank you for your suggestion. We have switched torch to a stable version. Please see the link https://github.com/buddy-compiler/buddy-mlir/pull/266 for details.

zhanghb97:
@BHbean Thanks for reporting this! We have fixed the issue. Please check😄

BHbean:
Thanks a lot! The `torch2.1.2+cpu` version did work!
> Besides, it seems that the `pybind11` package needs to be installed extraly by running `conda install pybind11 -c conda-forge` before building llvm and buddy-mlir, which will not be installed by running `pip install -r requirements.txt`.
Should this be added under `pip install -r requirements.txt` in `examples/BuddyLlama`? I could open a pr if you guys think it neccessary :)

zhanghb97:
@BHbean, feel free to add this! Thanks!

PR #272, opened 2024-01-02
add steps for installing 'pybind11' in examples/BuddyLlama/README.md

BHbean:
The BuddyLlama example requires `pybind11` package, which is not presented in `example/BuddyLlama/README.md`, so a supplement is made. ( #265 )


zhanghb97:
@BHbean Thank you for adding this! I think adding the pybind11 dependency to [requirements.txt](https://github.com/buddy-compiler/buddy-mlir/blob/main/requirements.txt) is a better choice. WDYT?

PR #279, opened 2024-01-26
[DIP] Modify imgcodecs and ImageContainer to update buddy-benchmark

meshtag:
This PR intends to help in the process of updating `buddy-benchmark` by marking related functions inline and applying other changes. This is done with the motive of resolving their definition reference in the single compilation unit created by linking multiple modular benchmarks in `buddy-benchmark`.


zhanghb97:
Thanks!

PR #282, opened 2024-01-30
[Midend]use declarative-pass-specification.

linuxlonelyeagle:


midend/include/Transforms/Passes.h
```diff
@@ -0,0 +1,21 @@
+#ifndef BUDDY_TRANSFORMS_PASSES_H
```
zhanghb97, line 1: Please add a license header.

midend/include/Transforms/Passes.td
```diff
@@ -0,0 +1,13 @@
+include "mlir/Pass/PassBase.td"
```
zhanghb97, line 1: Please add a license header.

linuxlonelyeagle:
@zhanghb97 If something is wrong, feel free to modify it.

linuxlonelyeagle:
I seem to have found a new problem, should we adjust the namespace of register,for example mlir::buddy::registerMatMulOptimizePass(), should we adjust it to buddy::registerMatMulOptimizePass()? I'm not quite sure, because our own dialect is actually defined directly in buddy's namespace, and this implies that the item buddy is on the same level as mlir.

PR #285, opened 2024-02-09
[BuddyGPU] Add GPU infrastructure and implements end-to-end inference

SForeKeeper:
This pull request does three following things, and is mainly for demonstration. It's better to split them into different PRs.

Adds two passes for replacing CPU allocations to GPU operations.
Make the GPU operations async by a dataflow analysis. (Still working on it)
Optimize linalg operators using transform dialect operations.
Apply the transformation process on an end-to-end inference module, then lower it for execution.
Notice: As this PR does not include the support for bf16 operations, it will not work on graphics cards with insufficient global memory. A newer operation gpu.dynamic_shared_memory is also used for lowering, yet it's not supported in Buddy's LLVM version.


carolove:
is there any progress in this PR? when will this pr being merged into main stream

Issue #164, opened 2023-06-13
[DIP] Create methods for splitting and merging channels in Image Container


meshtag:
**Is your feature request related to a problem? Please describe.**
As of now, we cannot work with individual channels directly in a multi-channel image stored in the image container. 

**Describe the solution you'd like**
We should have separate methods for splitting and merging channels from/in the image container. Similar to `cv2.split()` and `cv2.merge()` for OpenCV's `Mat` container.

zhanghb97:
@meshtag Thanks!
ping @NK-MXD @FlagerLee

FlagerLee:
I'm currently busy on other project and not involved in development of DIP. @meshtag

zhanghb97:
@NK-MXD and @FlagerLee have not continued working in this direction.

PR #288, opened 2024-02-25
[DIP] Fix DIP codegen for supporting multi-channel processing

meshtag:
This PR handles the codegen backend part. 
Comprehensive fix to https://github.com/buddy-compiler/buddy-mlir/issues/164 is the end-goal.


meshtag:
[This](https://gist.github.com/meshtag/35c1a98e53cbaba71daf671a52ba24d1) is a simple sanity test for verifying that multi-channel processing is working as expected. It can be tested by lowering the file and then using `mlic-cpu-runner`.

The lowering pass pipeline is as follows: 
```
-convert-linalg-to-loops -lower-affine -convert-scf-to-cf -convert-scf-to-cf -convert-vector-to-llvm --finalize-memref-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts
```

PR #289, opened 2024-02-28
[Gemmini] Refactor lower-gemmini pass

qzylalala:
1. We spilt `lower-gemmini` pass to two individual passes `lower-gemmini` and `lower-gemmini-to-llvm`.
2. The new `lower-gemmini` pass will lower gemmini operations to 3rd library(libgemmini) function call.
3. We also remain the code for old `lower-gemmini` pass, you can choose them using  populate function in `midend/include/Dialect/Gemmini/Transform.h`.

- `populateGemminiLegalizeForLLVMExportPatterns()` for the old one.
- `populateGemminiLegalizeForFuncExportPatterns()` for the new one.


linuxlonelyeagle:
@qzylalala  Why does `libgemmini` exist? What is the benefit of this? What is the benefit of using assembly? If assembly is generated like this, then why write a compiler?The `gemmini intrinsic op` is at the same level as the op in the `llvm dialect`.So splitting into two passes I think is very strange.

qzylalala:
> @qzylalala Why does `libgemmini` exist? What is the benefit of this? What is the benefit of using assembly? If assembly is generated like this, then why write a compiler?The `gemmini intrinsic op` is at the same level as the op in the `llvm dialect`.So splitting into two passes I think is very strange.

I agree with your opinion. However, I need to clarify the current new requirement.  Hongbin hopes that we can try to use the method of calling third-party libraries to complete the compilation（Another team need this feature）. Therefore, before transforming the gemmini operator into intrinsic op, I replace this process with a pass that calls third-party functions.

linuxlonelyeagle:
@qzylalala I recommend you rewrite a pass.

linuxlonelyeagle:
> @qzylalala I recommend you rewrite a pass.

Rewrite a pass for your situation.

linuxlonelyeagle (commented):
The most important point is that no changes should be made to the existing lower-gemmini.My suggestion, either remove the backend and delete the existing lower-gemmini, or don't change the current lower-gemmini and rewrite the pass as needed.

linuxlonelyeagle:
@zhanghb97  @qzylalala I think we need to have a discussion here. To give you my thoughts, if we do this as we do now, I think we no longer need to use the existing backend, and of course in doing so I don't think there is any need for the lower-gemmini pass to exist any more, which I think is actually acceptable because the backend does cause a lot of problems. If the backend still exists, then no changes should be made to the existing lower-gemmini, and in this PR, the matching pattern of lower-gemmini is changed.Outside of the first time, there are a lot of problems with this PR, compilation failure which is a problem, no endline in the code, and a lot of @qzylalala local computer with some settings.I think this PR needs to be stopped first, and there should be a discussion here before we can move on to the next step.

qzylalala:
> The most important point is that no changes should be made to the existing lower-gemmini.My suggestion, either remove the backend and delete the existing lower-gemmini, or don't change the current lower-gemmini and rewrite the pass as needed.

I also tend to not modify the original code, but to rewrite a new pass. In this way, the makefile does not need to be modified, and testing can also pass. After completing the functional testing, I will rearrange the project structure.

PR #291, opened 2024-03-19
fix lenet error and format files

weilinquan:


weilinquan:
@zhanghb97 I still lack test file for func-bufferize-dynamic-offset. Is there a example for this? Should I refer tests/Conversion/matmul-vectorization.mlir?

zhanghb97:
Thanks!

PR #293, opened 2024-04-17
add graph_json module

wdjyd:
Added a method to the `Graph` class for converting the graph to JSON format.
Added a method to the `Graph` class for converting the graph to DOT format for visualization.
Added a method in the "json_decoder.py" for converting from a JSON string to a Graph object.

The `json_to_graph` method has not been merged into the `Graph` class temporarily due to circular import issues.

Usage examples for the changes have been added to the BuddyLeNet model.


xlinsist:
Recommend adding usage examples(e.g. how to convert .dot to .svg) to README under the BuddyLeNet directory

PR #298, opened 2024-05-01
ci: bump cache & checkout version 4

c8ef:
This PR updates the cache and checkout action to version 4. There are no functional changes.


zhanghb97:
Thanks!

PR #300, opened 2024-05-08
[Gemmini] Fuse

qzylalala:


midend/lib/Conversion/LowerLinalgToGemmini/LowerLinalgToGemmini.cpp
```diff
@@ -63,10 +63,42 @@ class MatmulLowering : public OpRewritePattern<linalg::MatmulOp> {
     Value fillOpInputValue =
         rewriter.create<arith::ConstantOp>(loc, fillOpInsType, fillOpInputAttr);
     rewriter.create<linalg::FillOp>(loc, fillOpInputValue, bias);
-    rewriter.replaceOpWithNewOp<gemmini::TileMatMulOp>(
-        matMulOp, input0, input1, output0, bias, /*aScaleFactor = */ scale1,
-        /*bScaleFactor = */ scale1, /*dScaleFactor = */ scale1, /*act = */ 0,
-        /*accScale = */ scale1, /*bertScale = */ scale0);
+
+    // llvm::outs() << " has "
```
linuxlonelyeagle, line 67: Clear the comments here.

midend/lib/Conversion/LowerLinalgToGemmini/LowerLinalgToGemmini.cpp
```diff
@@ -63,10 +63,42 @@ class MatmulLowering : public OpRewritePattern<linalg::MatmulOp> {
...
+    //                                 output0.getUses().end())
+    //                << " uses:\n";
+    // for (Operation *userOp : output0.getUsers()) {
+    //   llvm::outs() << "    - " << userOp->getName() << "\n";
+    // }
+
+    // If this matmul operation is followed by a transpose operation, do fusion.
+    // We should make sure that the result of this matmul op only has one user.
+    Operation* fuseOp = *output0.user_begin();
+    int output0Use = 0;
```
linuxlonelyeagle, line 78: There is a serious error in logic here.I would suggest writing the pattern inside the canonicalizer of the `gemmini.tile_matmul` op.What you should be determining is whether there is only one user of `gemmini.tile_matmul` and it is `linalg.transpose`.`bTranspose` is a attribute, so  let it be the opposite.You shouldn't be judging the number of `linalg.transpose`, it's completely wrong to do so.A pass is based on a greedy algorithm, so this pass will continue to match against the pattern.

midend/lib/Conversion/LowerLinalgToGemmini/LowerLinalgToGemmini.cpp
```diff
@@ -63,10 +63,42 @@ class MatmulLowering : public OpRewritePattern<linalg::MatmulOp> {
...
+    for (auto userOp : output0.getUsers()) {
+      if (auto transposeOp = dyn_cast<linalg::TransposeOp>(userOp)) {
+        fuseOp = transposeOp;
+        output0Use ++;
+      }
+    }
+
+    if (output0Use) {
+      // llvm::outs() << "Fuse linalg.matmul and linalg.transpose. \n";
```
linuxlonelyeagle, line 87: Remove comments

linuxlonelyeagle (changes requested):
There are a lot of problems here.

linuxlonelyeagle:
Regarding the conclusion that transpose is not supported, it is recommended that you write a C program to test it. The table is an indication of whether or not to transpose A and B. In this case, it is transpose C.

midend/lib/Conversion/LowerLinalgToGemmini/LowerLinalgToGemmini.cpp
```diff
@@ -63,10 +63,42 @@ class MatmulLowering : public OpRewritePattern<linalg::MatmulOp> {
...
+      rewriter.replaceOpWithNewOp<gemmini::TileMatMulOp>(
+            matMulOp, input1, input0, output0, bias, /*aScaleFactor = */ scale1,
+            /*bScaleFactor = */ scale1, /*dScaleFactor = */ scale1, /*act = */0,
+            /*accScale = */ scale1, /*bertScale = */ scale0,
+            /*repeatingBias = */ false, /*aTranspose = */ true, 
+            /*bTranspose = */ true);
```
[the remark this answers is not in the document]
qzylalala: For `C = AB, C = C^T`, we use this formula : `C^T = (AB)^T = B^T A^T`.
So, we change the position of `input0` and `input1`, and set `aTranspose` and `bTranspose` true.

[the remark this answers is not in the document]
linuxlonelyeagle: If so, the reasoning is the same, you just need to make `aTranspose` the opposite of `bTransepose` as before.

midend/lib/Conversion/LowerLinalgToGemmini/LowerLinalgToGemmini.cpp
qzylalala: Hi, thanks for your suggestions.  I agree that it is indeed possible to do the match during `gemmini.tile_matmul` lowering. However, I still don’t understand why there is a problem with greedy matching here. Can you give an example here? Maybe I’m not that familiar with it yet.

linuxlonelyeagle:
```
%1 = gemmini.tile_matmul  {transepose = false}
.....
%4 = linalg.transpose %1 
.....
% 7 = linalg.transpose %4
.....
use(%7)
```
first
```
%1 = gemmini.tile_matmul  {transepose = true}
.....
% 7 = linalg.transpose %1
.....
use(%7)
```
second
```
%1 = gemmini.tile_matmul {transepose = false}
.....
use(%1)
```

linuxlonelyeagle:
```
%1 = gemmini.tile_matmul  {transepose = false}
.....
%4 = linalg.transpose %1 
.....
%7 = linalg.transpose %1
```
I'm not sure if this happens in the model not sure if it does.
So for this fuse pattern, you should make sure that there is only one op that uses `gemmini.tile_matmul` and that op is  `linalg.transpose`.

Issue #115, opened 2023-02-11
[Testing] Improve Image Container tests


meshtag:
**Is your feature request related to a problem? Please describe.**
[Current tests](https://github.com/buddy-compiler/buddy-mlir/blob/main/tests/Interface/core/ImageContainerTest.cpp) for the `image container` do not test the `normalization` functionality and the accuracy of stored pixel value when directly compared with the original image for all pixels. 

**Describe the solution you'd like**
Add simple tests to check whether the stored values in wrapped `memref` (`Img` container) is in agreement with the original image pixel values (with and without `norm` flag).

aPR0T0:
I would like to work on this issue, taking a quick look into the source code

markram1729:
Hi! , I have looked into buddy-mlir Interface/DIP code is this issue still open ?

meshtag:
> Hi! , I have looked into buddy-mlir Interface/DIP code . I would like to work on this issue,is this issue still open ?

Hey, thanks for your interest. The issue is still open and contributions are welcome.

markram1729:
Hey , are there any improvements  needed for the image container , i see issue is not closed yet

zhanghb97:
We now have a new implementation and [test cases](https://github.com/buddy-compiler/buddy-mlir/blob/main/tests/Interface/core/NewImageContainerTestBmp.cpp) of the container that allow us to eliminate the dependency on OpenCV. It might be a good idea to expand the testing scope.

PR #303, opened 2024-05-15
opencv test draft for ImageContainer

markram1729:
Draft PR for opencv test integration #115

### DIFF
--- a/tests/Interface/core/ImageContainerTest.cpp
+++ b/tests/Interface/core/ImageContainerTest.cpp
@@ -24,6 +24,14 @@
 #include <buddy/Core/Container.h>
 #include <buddy/DIP/ImageContainer.h>
 
+bool compare_flt(float a,float b)
+{
+	if(a==b) return true;
+        float eps = a-b;
+        if(eps<0)eps=-eps;
+        if(eps<FLT_EPSILON) return true;
+        return false;
+}
 template<typename T,size_t N>
 bool testcvdp(cv::Mat testImgcv,Img<T,N> testImg,bool norm = false)
 {
@@ -41,7 +49,14 @@ bool testcvdp(cv::Mat testImgcv,Img<T,N> testImg,bool norm = false)
 		{
 			for(size_t j=0;j<testImg.getSizes()[1];j++)
 			{
-				if(data[k]!=(T)testImgcv.at<uchar>(i,j)) return false;
+				if(!norm)
+				{
+					if(!compare_flt(data[k],(T)testImgcv.at<uchar>(i,j))) return false;
+				}
+				else
+				{
+					if(!compare_flt(data[k],(T)testImgcv.at<T>(i,j)))return false;
+				}
 				k++;
 			}
 		}
@@ -167,6 +182,15 @@ int main() {
   //CHECK: 1
   fprintf(stderr,"%d \n",testbmp);
   //===--------------------------------------------------------------------===//
+  // Test Opencv Image with norm
+  //===--------------------------------------------------------------------===//
+  Img<float,2> testImgbmpnorm(testImgcvbmp,nullptr,true);
+  cv::Mat checkimgbmp(testImgcvbmp.rows,testImgcvbmp.cols,CV_32FC1);
+  testImgcvbmp.convertTo(checkimgbmp,CV_32FC1,1.f/255);
+  bool testbmp1=testcvdp<float,2>(checkimgbmp,testImgbmpnorm,true);
+  //CHECK: 1
+  fprintf(stderr,"%d \n",testbmp1);
+  //===--------------------------------------------------------------------===//
   // Test jpeg format image.
   //===--------------------------------------------------------------------===//
   Img<float, 2> grayimage_jpg = dip::imread<float, 2>(
@@ -274,7 +298,13 @@ int main() {
   bool testjpg=testcvdp<float,2>(testImgcvjpg,testImgjpg);
   //CHECK: 1
   fprintf(stderr,"%d \n",testjpg);
-
+  //===--------------------------------------------------------------------===//
+  // Test Opencv Image with norm
+  //===--------------------------------------------------------------------===//
+  Img<float,2> testImgjpgnorm(testImgcvjpg,nullptr,true);
+  cv::Mat checkimgjpg(testImgcvjpg.rows,testImgcvjpg.cols,CV_32FC1);
+  testImgcvjpg.convertTo(checkimgjpg,CV_32FC1,1.f/255);
+  bool testjpg1=testcvdp<float,2>(checkimgjpg,testImgjpgnorm,true);
   //===--------------------------------------------------------------------===//
   // Test png format image.
   //===--------------------------------------------------------------------===//
@@ -373,7 +403,9 @@ int main() {
   const Img<float, 2> testBracketOperator6(grayimage_png);
   // CHECK: 240.0
   fprintf(stderr, "%f\n", testBracketOperator6[15]);
-
+  //===--------------------------------------------------------------------===//
+  // Test Opencv Image without norm
+  //===--------------------------------------------------------------------===//
    cv::Mat testImgcvpng=cv::imread(
                   "../../../../tests/Interface/core/TestGrayImage.png",
                   cv::IMREAD_GRAYSCALE);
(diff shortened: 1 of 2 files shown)


tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,45 @@
...
+bool compare_flt(float a,float b)
+{
+	if(a==b) return true;
+        float eps = a-b;
+        if(eps<0)eps=-eps;
+        if(eps<FLT_EPSILON) return true;
+        return false;
+}
+template<typename T,size_t N>
+bool testcvdp(cv::Mat testImgcv,Img<T,N> testImg,bool norm = false)
```
meshtag, line 36: Can you use a more descriptive name here?

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,45 @@
...
+	if(N==2)
+	{
+		int k=0;
+		for(size_t i=0;i<testImg.getSizes()[0];i++)
+		{
+			for(size_t j=0;j<testImg.getSizes()[1];j++)
+			{
+				if(!norm)
+				{
+					if(!compare_flt(data[k],(T)testImgcv.at<uchar>(i,j))) return false;
```
meshtag, line 54: No need to explicitly return false here, just return the function result directly.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,45 @@
...
+				}
+				else
+				{
+					if(!compare_flt(data[k],(T)testImgcv.at<T>(i,j)))return false;
```
meshtag, line 58: Both of these branches (if and else block) can be simplified using a ternary operator.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,45 @@
...
+	int cvn = testImgcv.dims;
+	if(cvn!=N) return false;
+	for(int i=0;i<N;i++)
+	{
+		if(testImgcv.size[i]!=testImg.getSizes()[i]) return false;
+	}
+	T *data= testImg.getData();
+	if(N==2)
+	{
+		int k=0;
```
meshtag, line 47: Do you really require signedness from the variable here? You can use unsigned in all cases where it is not required.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,45 @@
 #include <buddy/Core/Container.h>
 #include <buddy/DIP/ImageContainer.h>
 
+bool compare_flt(float a,float b)
+{
+	if(a==b) return true;
+        float eps = a-b;
+        if(eps<0)eps=-eps;
```
meshtag, line 31: You can use `std::abs` to eliminate an unnecessary branch here.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,45 @@
...
+        if(eps<FLT_EPSILON) return true;
```
meshtag, line 32: No need for explicit true and false here, you can just return the comparison result itself.

meshtag (changes requested):
Thanks for the work! 

I have added some high level comments.
Additionally, 
1. Can you please format everything with clang-format.
2. It is useful to have a descriptive summary of what the function intends to do on top of its definition/declaration.
3. It is a good idea to separate logically different lines of code with a newline for better readability.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,38 @@
 #include <buddy/Core/Container.h>
 #include <buddy/DIP/ImageContainer.h>
 
+bool compare_flt(float a, float b) {
+  if (a == b)
```
meshtag, line 28: No need to have this separate control flow branch here. Please handle it generically with below logic itself.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,38 @@
...
+    return true;
+  float eps = std::abs(a - b);
```
meshtag, line 30: No need to store the diff in a separate variable here.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,38 @@
...
+  return (eps < FLT_EPSILON);
+}
+
+template <typename T, size_t N>
+bool testImgcvnorm(cv::Mat testImgcv, Img<T, N> testImg, bool norm = false) {
+  int cvn = testImgcv.dims;
+  if (cvn != N)
+    return false;
+  for (size_t i = 0; i < N; i++) {
```
meshtag, line 39: Nit: Using pre-increment (++i instead of i++) is marginally better wherever you can (pls modify other places as well).

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,38 @@
...
+    if (testImgcv.size[i] != testImg.getSizes()[i])
+      return false;
+  }
+  T *data = testImg.getData();
+  if (N == 2) {
+    size_t k = 0;
+    for (int i = 0; i < testImg.getSizes()[0]; i++) {
+      for (int j = 0; j < testImg.getSizes()[1]; j++) {
+        if ((norm ? !compare_flt(data[k], (T)testImgcv.at<T>(i, j))
```
meshtag, line 48: Extra set of parentheses here?

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -132,7 +164,25 @@ int main() {
...
   fprintf(stderr, "%f\n", testBracketOperator2[15]);
-
+  //===--------------------------------------------------------------------===//
+  // Test Opencv Image without norm
+  //===--------------------------------------------------------------------===//
+  cv::Mat testImgcvbmp=cv::imread(
+		  "../../../../tests/Interface/core/TestGrayImage.bmp",
+		  cv::IMREAD_GRAYSCALE);
+  Img<float,2> testImgbmp(testImgcvbmp);
+  bool testbmp=testImgcvnorm<float,2>(testImgcvbmp,testImgbmp);
```
meshtag, line 174: Did you format this file with clang-format?

`git clang-format HEAD~` will format your entire previous commit in-place for you.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,38 @@
...
+  if (cvn != N)
+    return false;
+  for (size_t i = 0; i < N; i++) {
+    if (testImgcv.size[i] != testImg.getSizes()[i])
+      return false;
+  }
+  T *data = testImg.getData();
+  if (N == 2) {
+    size_t k = 0;
+    for (int i = 0; i < testImg.getSizes()[0]; i++) {
```
meshtag, line 46: These loops have redundant braces.

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -24,6 +24,38 @@
...
+bool testImgcvnorm(cv::Mat testImgcv, Img<T, N> testImg, bool norm = false) {
+  int cvn = testImgcv.dims;
+  if (cvn != N)
+    return false;
+  for (size_t i = 0; i < N; i++) {
+    if (testImgcv.size[i] != testImg.getSizes()[i])
+      return false;
+  }
+  T *data = testImg.getData();
+  if (N == 2) {
```
meshtag, line 44: Can you make this more generic instead of only handling for 2 dimensions?

meshtag (commented):
This is looking good. I have added few more comments.

tests/Interface/core/ImageContainerTest.cpp
markram1729: can you more give more information , generic as in for N>3 
but as i see at Imagecontainer.h ,for (N!=2)  cv image is not handled when N==3 / for NCHW layout i see copying only first channel information of image than number of channels 

can you give me direction /help on what can be done

meshtag: Can you support testing with the NCHW layout? (and add a TODO here to note that tests should be updated there once more layouts are supported functionally)

markram1729: can you please check now

markram1729:
Can @meshtag  @zhanghb97  review my pull request, waiting for your suggestions

tests/Interface/core/ImageContainerTest.cpp
```diff
@@ -332,6 +469,52 @@ int main() {
   const Img<float, 2> testBracketOperator6(grayimage_png);
   // CHECK: 240.0
   fprintf(stderr, "%f\n", testBracketOperator6[15]);
-  
+
+  //===--------------------------------------------------------------------===//
```
meshtag, line 473: These tests are testing the same thing which was tested by above tests? If not, please add a comment explaining that.

meshtag:
> Can @meshtag @zhanghb97 review my pull request, waiting for your suggestions

Sorry, I missed this. Can you please format all files with clang-format. 

Commented a minor suggestion, rest looks fine to me. 

Please wait for the approval of the person maintaining this part currently. Thanks!

tests/Interface/core/ImageContainerTest.cpp
markram1729: yes , above tests checks only on single images , here i have batched two images into a blob that is the only difference 

and i have added few required  segments in ImageContainer.h for batched images

zhanghb97:
Thanks!

PR #313, opened 2024-05-24
[DAP/Whisper]Add initial documents for Whisper Preprocessor.

taiqzheng:
Currently, it is in draft form with some remaining work to be completed.
However, it is now capable of being used to build an end-to-end Whisper model.

Work should be done in this PR:
- [x]  Correct format for data
The` clang-format `tool offers different strategies for formatting data. Please check file `ExtandDAPPass.cpp` in “Files changed”.
- [x] Add an example
The example has been designed, but the file location needs to be changed.
- [x] Correct the relationship with DAP Pass
Update file locations for the `extand-dap` pass and also change the operation name.

Future work, not in this PR:
- [x]  Update pass pipeline
The `one-shot-bufferize` pass has conflicts with some operations.
- [ ]  Update example
Currently only one case is supported, where its audio data is directly written into the example file. To support more audio cases, we need an audio library to read data from audio files. 
- [ ]  Extract operators
Some operators need to be extracted so that they can be reused in other algorithms.
- [ ] Optimize the algorithm
After extracting the operators, design vectorization pattern for each one.


frontend/Interfaces/lib/DAP-extend.mlir
```diff
@@ -0,0 +1,4 @@
+func.func @buddy_whisperPreprocess(%in : memref<?xf64>, %out : memref<1x80x3000xf32>) -> () {
+  dap.whisperPreprocess %in, %out : memref<?xf64>, memref<1x80x3000xf32>
```
zhanghb97, line 2: `dap.whisperPreprocess` -> `dap.whisper_preprocess`

zhanghb97 (approved):
Thanks!

PR #337, opened 2024-07-02
annotate the llama mlir code

effrey-liu:


zhanghb97:
Thanks! We will iteratively analyze the entire llama model based on this.

examples/BuddyLlama/llama_annotation.mlir
LHY-24, line 1: Good job! I want to make sure if the link of the original Llama model you used here is https://github.com/meta-llama/llama?

zhanghb97: @LHY-24 Nice catch! I noticed that the link in our [doc](https://github.com/buddy-compiler/buddy-mlir/tree/main/examples/BuddyLlama#buddy-compiler-llama-example) for llama inference now directs to llama3. We may need to update our llama example README and test the llama3 inference.

Also, @effrey-liu where is your original llama model from?

effrey-liu: Model comes from [meta.ai](https://llama.meta.com/), but I chose llama2-7B to generate mlir.

PR #342, opened 2024-07-08
[NFC] Make buddy tests and examples depends on `mlir-cpu-runner`

imkiva:
When `ninja check-buddy` is called without any prior `ninja check-mlir`, the following test failure is raised:

```
<...>/linalg-transpose-f32.mlir.script: line 1: <...>/buddy-mlir/build/./bin/mlir-cpu-runner: No such file or director
```

This PR fixes this by explicitly setting `mlir-cpu-runner` as a dependency of buddy tests and examples.


zhanghb97 (approved):
Thanks!

PR #344, opened 2024-07-10
add optimization plans for concat matmul, eliminate add-zero, eliminate useless identity op

effrey-liu:
add optimization plans for concat matmul, eliminate add-zero, eliminate useless identity op.


zhanghb97 (approved):
Thanks!

PR #357, opened 2024-07-29
frontend: add missing dependencies into DAP target

Avimitin:
Ninja introduced different build graph algorithm in version v1.12.0, which will required project to explicitly specify dependencies. Without this commit, ninja v1.12.0 will not link mlir-translate and llc before the DAP target and causing build failure.


zhanghb97 (approved):
Thanks!

PR #359, opened 2024-08-07
[build system] use variable from LLVMConfig

Avimitin:
In the previous build script, we always assumed that the binary and the library are relative to the MLIR_DIR variable. But this prevents buddy-mlir from properly packaging by the system. We should use the LLVMConfig.cmake file to have a more flexible build system.

After this PR, user can choose from building LLVM in submodules or using LLVM in their system.


CMakeLists.txt
```diff
@@ -42,13 +42,11 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR OR BUDDY_MLIR_OUT_OF_TREE_
     message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
     message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 
-    set(LLVM_MLIR_BINARY_DIR ${MLIR_DIR}/../../../bin)
-    set(LLVM_MLIR_LIBRARY_DIR ${MLIR_DIR}/../../../lib)
-    set(LLVM_PROJECT_BUILD_DIR ${MLIR_DIR}/../../../)
-    if(NOT DEFINED LLVM_PROJECT_SOURCE_DIR)
-      get_filename_component(LLVM_PROJECT_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/llvm/ ABSOLUTE)
+    # LLVM_MAIN_SRC_DIR only set on LLVM build. Here is an compatible layer for old build strategy.
```
zhanghb97, line 45: The comments here are a bit difficult to understand. What do "compatible layer" and "old build strategy" mean respectively? If I understand correctly, if users want to use LLVM in their own systems, they need to specify the LLVM directory through `LLVM_MAIN_SRC_DIR` during configuration, right? Then we may need to give an explanation in the README.

zhanghb97 (commented):
LGTM!
We need to clearly describe how to use the CMake configuration (LLVM_MAIN_SRC_DIR) in the appropriate place(top README file).

CMakeLists.txt
Avimitin: For users that want to use LLVM on their system, they can set this variable themselves. For people that want to use the LLVM inside the buddy-mlir git tree, they can just left this untouched. I will add a new guide for the former build strategy in READEME.

Avimitin:
Document and comments are updated.

PR #366, opened 2024-08-13
[Audio Container] Add encoder for wav audio format.

taiqzheng:
**Deviation:**
There might be slight discrepancies in the sample values after encoding, with a difference of one when represented as bytes (uint8_t).

**Example:**
Raw audio data:  fff7 
Encoded result: fff8

### DIFF
--- a/frontend/Interfaces/buddy/DAP/AudioContainer.h
+++ b/frontend/Interfaces/buddy/DAP/AudioContainer.h
@@ -87,7 +87,7 @@ private:
   int32_t fourBytesToI32(const std::vector<uint8_t> &fileData,
                          size_t startIndex,
                          Endianness endianness = Endianness::LittleEndian);
-  // Convert four bytes to a 16-bit integer according to byte order of data.
+  // Convert two bytes to a 16-bit integer according to byte order of data.
   int16_t twoBytesToI16(const std::vector<uint8_t> &fileData, size_t startIndex,
                         Endianness endianness = Endianness::LittleEndian);
   // Normalize 8-bit unsigned integer sample to a range of -1.0 to 1.0.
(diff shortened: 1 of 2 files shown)


frontend/Interfaces/buddy/DAP/AudioContainer.h
```diff
@@ -142,6 +167,49 @@ template <typename T, std::size_t N> Audio<T, N>::Audio(std::string filePath) {
...
+  // Convert the string to lowercase before comparison, ensuring that case
+  // variations are handled without repeating conditions.
+  std::transform(format.begin(), format.end(), format.begin(), ::tolower);
+  // Vector for storing bytes in a specific audio format.
+  std::vector<uint8_t> fileData;
+  // Select encoder.
+  if (format == "wav" || format == "wave") {
+    bool success = EncodeWaveFile(fileData);
+    if (!success) {
+      this->audioFormat = AudioFormat::ERROR;
```
zhanghb97, line 186: Do not need to modify the format.

frontend/Interfaces/buddy/DAP/AudioContainer.h
```diff
@@ -142,6 +167,49 @@ template <typename T, std::size_t N> Audio<T, N>::Audio(std::string filePath) {
...
+      throw std::runtime_error("Failed to encode WAVE file.");
```
zhanghb97, line 187: Using the following error report method:
```
std::cerr << Failed to encode WAVE file. << std::endl;
return false;
```

tests/Interface/core/AudioContainerTest.cpp
```diff
@@ -20,13 +20,20 @@
...
 #include <iostream>
 
 using namespace std;
 
 int main() {
+  // ---------------------------------------------------------------------------
+  // 1. Print Decoded Reuslts using Buddy Audio Container
+  // ---------------------------------------------------------------------------
+
+  // Read and decode audio file with Buddy Audio Contaier.
```
zhanghb97, line 34: Contaier -> Container

PR #368, opened 2024-08-15
[feat] Rotate4D

hharryz:
Finish Rotate4D Operation in DIP dialect.

1. [feat] slightly accelerate rotate2d (rearrange memory layout)
2. [bugfix] fix the bug of rotate2d (create right way of padding)
3. [feat] finish rotate4d and do acceleration, now supporting nhwc & nchw format
4. [feat] add rotate4d example

run commands (in buddy-mlir dir)
$ cd build
$ cmake -G Ninja .. -DBUDDY_EXAMPLES=ON -DBUDDY_ENABLE_OPENCV=ON -DBUDDY_MLIR_ENABLE_DIP_LIB=ON
$ ninja rotation4D
$ cd bin
$ ./rotation4D ../../examples/images/YuTu.png result-dip-rotate.png


examples/DIPDialect/rotation4D.cpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return 1;
+}
+
+int main(int argc, char *argv[]) {
+  testImplementation(argc, argv);
+
+  return 0;
+}
```
zhanghb97, line 104: Please add an empty line here.

zhanghb97 (commented):
Giving quick feedback. Please add unit test for your operation and vectorization pass.

hharryz:
The following function will cause linker failure if the DIP related header files are included multiple times.
an `inline` notation has to be added to avoid redefinition!
`inline void FillPalette(PaletteBlock *palette, int bpp, bool negative = false); `

PR #369, opened 2024-08-16
[Examples] Adapt Audio Container for dap examples.

taiqzheng:
1.  Adapt new audio container for the following examples.
- [x] buddy-whisper-preprocess
- [x] buddy-fir
- [x] buddy-biquad
- [x] buddy-iir-scalar
- [x] buddy-iir-vectorization
2.  Merge 'BuddyLibDAPVectorization' library into 'BuddyLibDAP' library.
> We have different pass for scalar op, vectorized op and extend op. Now, all in one library. 

3.  Update the new Audio Contaienr to support the conversion from a MemRef object to an Audio object.
> A move constructor has been designed. 

4.  Handle corner case for Audio encoder. 
> NaN/-NaN will be reset to 1. (same with KFR library)


zhanghb97 (approved):
Thanks!

PR #371, opened 2024-08-19
[feature] new img container support BMP 24-16 bit

Hanyonggong:
New img container support BMP 24-16 bit
* suport BMP 24 bit
* suport BMP 16 bit (two format：RGB565 and RGB 555)

All have passed the Lenet model test


LiuYang328:
I test 32, 24, 16 bit  BMP in MobileNet example, it can compiled but the result of the calculation is wrong.
If i use opencv version container,  the input is "dog.png", the result is "Samoyed", this is right, but the new container results "drake"
image

Hanyonggong:
@LiuYang328 This may be not a problem with the new container, maybe it's because no resizing after using new container

LiuYang328:
> @LiuYang328 This may be not a problem with the new container, maybe it's because no resizing after using new container

The input is already resized, the bmp file both 224*224

Hanyonggong:
> > @LiuYang328 This may be not a problem with the new container, maybe it's because no resizing after using new container
> 
> The input is already resized, the bmp file both 224*224

You can compare the differences between them, such as shape, it can work normally on my end

LiuYang328 (approved):

zhanghb97 (approved):

PR #374, opened 2024-08-20
[Vectorization/pooling] add Pooling Nhwc Max Vectorization pass and relevant case

FloatingcloudKnight:
add Pooling Nhwc Max Vectorization pass and relevant case


examples/MLIRLinalg/linalg-pooling-nhwc-max.mlir
```diff
@@ -0,0 +1,50 @@
+// RUN: buddy-opt %s \
+// RUN:     -convert-linalg-to-loops -lower-affine -convert-scf-to-cf \
```
zhanghb97, line 2: Please align this part of the code.

PR #382, opened 2024-09-03
Add MLIR files and op for three models in Stable Diffusion

R-Tars:
Add MLIR files and op for three models in Stable Diffusion


zhanghb97 (commented):
Please use `black` to format all the python files.

zhanghb97 (approved):

zhanghb97:
Thanks!

PR #398, opened 2024-10-10
[Example] Modify container test and libpng cmakelists.

LiuYang328:


tests/Interface/core/NewImageContainerTest.cpp
```diff
@@ -28,7 +28,8 @@ int main() {
   //===--------------------------------------------------------------------===//
   // Default Gray Scale
   dip::Image<float, 4> bmp32bitGrayDefault(
-      "../../../../tests/Interface/core/TestImage.bmp", dip::DIP_GRAYSCALE);
+      "../../../../tests/Interface/core/TestImage-gray.bmp",
```
LiuYang328, line 31: after clang-format -i

PR #407, opened 2024-10-16
[BuddyMobileNetV3]Remove the num_batches_tracked attribute

WuXintong123:
- Remove the num_batches_tracked attribute.
- Removed the int64 type parameter file arg1.data


zhanghb97 (approved):
Thanks!

PR #412, opened 2024-10-19
[DIP] add bmp encode to write image

Hanyonggong:
add api: dip::imageWrite


LiuYang328:
Usage :
 ./resize4D_nhwc ../../examples/images/YuTu.png result-dip-resize.bmp

LiuYang328:
The function is correct ! 
The code format needs to be changed.
I think this could be written as a separate example .cpp file, only test if the image is generated correctly

zhanghb97 (approved):
@Hanyonggong @LiuYang328 Thanks!

PR #415, opened 2024-10-21
[GPU] Add an example of running LeNet on GPU.

NaOHCC:


zhanghb97 (commented):
- Please update README for the BuddyLeNet example.
- I got different results from the CPU and GPU targets.
image

NaOHCC:
Corrected the result.

zhanghb97 (approved):
Thanks!

PR #442, opened 2024-12-30
[examples] Add several examples of `vector` dialect on GPU

xTayEx:


zhanghb97:
Currently, our CI server does not have a GPU, so add the `MLIRVectorGPU` path to [this list](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/lit.cfg.py#L38) to avoid test failures.

xTayEx:
> Currently, our CI server does not have a GPU, so add the `MLIRVectorGPU` path to [this list](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/lit.cfg.py#L38) to avoid test failures.

I've excluded the `MLIRVectorGPU` path and now the CI passes. Please review it when you have time. Thanks!

examples/MLIRVectorGPU/makefile
```diff
@@ -0,0 +1,1022 @@
+#!/bin/bash
+BUDDY_OPT := ../../build/bin/buddy-opt
```
zhanghb97, line 2: Please use the variable definition like [this](https://github.com/buddy-compiler/buddy-mlir/blob/main/examples/BuddyMatmul/makefile#L2) in case of modifications when changing the `build` path.

examples/MLIRVectorGPU/vector-bitcast.mlir
```diff
@@ -0,0 +1,45 @@
...
+    call @printMemrefF32(%kernel_ret1_cast) : (memref<*xf32>) -> ()
+    call @printMemrefI32(%kernel_ret2_cast) : (memref<*xi32>) -> ()
+
+    func.return
+  }
+  func.func private @printMemrefI64(%tpr : memref<*xi64>)
+  func.func private @printMemrefF32(%ptr : memref<*xf32>)
+  func.func private @printMemrefI32(%ptr : memref<*xi32>)
+
+}
```
zhanghb97, line 45: Please add an empty line here (and the following files).

zhanghb97:
Thanks!

PR #463, opened 2025-02-18
[midend/lib/Conversion/LowerLinalgToGemmini] Add pass support for gemmini to run E2E LeNet and some tests.

shirohasuki:
[midend/lib/Conversion/LowerLinalgToGemmini] Add conv2d_nhwc_fhwc pass for gemmini and add relevant examples and tests.
[midend/lib/Conversion/LowerGemmini] Add scalar print Op for convenient testing and add relevant examples and tests.
[midend/lib/Dialect/Gemmini] Split gemmini conv's RISC and CISC instruction lowering into two separate passes. (Now it's default lowering to RISC type different from native-gemmini, maybe we need a option to config this. Both can pass the LENet test.)

Below are the RISC version and CISC version running LeNet E2E on firesim.
`L%M6F8O34OTEQ39G8FM`)5

### DIFF
--- a/examples/GemminiDialect/conv_2d_nhwc_fhwc_5x5_i8.mlir
+++ b/examples/GemminiDialect/conv_2d_nhwc_fhwc_5x5_i8.mlir
@@ -1,0 +1,32 @@
+// RUN: buddy-opt %s \
+// RUN:     --convert-linalg-to-gemmini | \
+// RUN: FileCheck %s
+
+memref.global "private" @input : memref<1x7x7x1xi8> = dense<[[[[1],[1],[1],[1],[1],[1],[1]],
+                                                              [[1],[1],[1],[1],[1],[1],[1]],
+                                                              [[1],[1],[1],[1],[1],[1],[1]],
+                                                              [[1],[1],[1],[1],[1],[1],[1]],
+                                                              [[1],[1],[1],[1],[1],[1],[1]],
+                                                              [[1],[1],[1],[1],[1],[1],[1]],
+                                                              [[1],[1],[1],[1],[1],[1],[1]]]]>
+
+memref.global "private" @kernel : memref<1x5x5x1xi8> = dense<[[[[1], [1], [1], [1], [1]],
+                                                               [[1], [1], [1], [1], [1]],
+                                                               [[1], [1], [1], [1], [1]],
+                                                               [[1], [1], [1], [1], [1]],
+                                                               [[1], [1], [1], [1], [1]]]]>
+
+func.func @main() -> i8 {
+  %0 = arith.constant 0 : i8
+  %input = memref.get_global @input : memref<1x7x7x1xi8>
+  %kernel = memref.get_global @kernel : memref<1x5x5x1xi8>
+  %output = memref.alloc() : memref<1x3x3x1xi8>
+
+  // CHECK: gemmini.tile_conv %{{[0-9]+}} %alloc_{{[0-9]+}} %alloc_{{[0-9]+}} %alloc_{{[0-9]+}} %{{.+}} %{{.+}} :
+  // CHECK-SAME: memref<1x7x7x1xi8> memref<25x1xi8> memref<1xi32> memref<9x1xi8> i64 i64
+  linalg.conv_2d_nhwc_fhwc
+    ins(%input, %kernel : memref<1x7x7x1xi8>, memref<1x5x5x1xi8>)
+  outs(%output : memref<1x3x3x1xi8>)
+  gemmini.print %output : memref<1x3x3x1xi8>
+  return %0 : i8
+}
--- a/midend/lib/Conversion/LowerGemmini/LowerGemminiPass.cpp
+++ b/midend/lib/Conversion/LowerGemmini/LowerGemminiPass.cpp
@@ -152,6 +152,95 @@ private:
   }
 };
 
+class PrintScalarOpLowering : public ConversionPattern {
+public:
+  explicit PrintScalarOpLowering(MLIRContext *context)
+      : ConversionPattern(gemmini::PrintScalarOp::getOperationName(), 1,
+                          context) {}
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto context = rewriter.getContext();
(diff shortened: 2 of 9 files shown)


shirohasuki:
The commits seem to be missing the endline? Although I can see them locally.....

examples/GemminiDialect/conv_2d_nhwc_fhwc_5x5_i8.mlir
```diff
@@ -0,0 +1,32 @@
...
+                                                               [[1], [1], [1], [1], [1]]]]>
+
+func.func @main() -> i8 {
+  %0 = arith.constant 0 : i8
+  %input = memref.get_global @input : memref<1x7x7x1xi8>
+  %kernel = memref.get_global @kernel : memref<1x5x5x1xi8>
+  %output = memref.alloc() : memref<1x3x3x1xi8>
+
+  // CHECK: gemmini.tile_conv %{{[0-9]+}} %alloc_{{[0-9]+}} %alloc_{{[0-9]+}} %alloc_{{[0-9]+}} %{{.+}} %{{.+}} :
+  // CHECK-SAME: memref<1x7x7x1xi8> memref<25x1xi8> memref<1xi32> memref<9x1xi8> i64 i64
```
linuxlonelyeagle, line 26: Maybe you don't need `check-same`, and feel that the content of `check-smae` is not important. The same is true for the following example.

linuxlonelyeagle: gemmini.tile_conv ({{.*}} ({{.*}} ...， Maybe better since you're not capturing the variable and then using it.

midend/lib/Conversion/LowerGemmini/LowerGemminiPass.cpp
```diff
@@ -152,6 +152,95 @@ class PrintOpLowering : public ConversionPattern {
...
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto context = rewriter.getContext();
+    auto loc = op->getLoc();
+
+    ModuleOp parentModule = op->getParentOfType<ModuleOp>();
+
+    auto printfRef = getOrInsertPrintf(rewriter, parentModule);
```
linuxlonelyeagle, line 169: Do not use `auto`,The data types here can be confusing for people who are not familiar with this code.

midend/lib/Conversion/LowerGemmini/LowerGemminiPass.cpp
```diff
@@ -152,6 +152,95 @@ class PrintOpLowering : public ConversionPattern {
...
+
+    Value valueToPrint = op->getOperand(0);
+    if (elementType == rewriter.getF32Type()) {
+      valueToPrint = rewriter.create<LLVM::FPExtOp>(loc, rewriter.getF64Type(),
+                                                    valueToPrint);
+    } else if (elementType == rewriter.getI8Type()) {
+      valueToPrint = rewriter.create<LLVM::SExtOp>(loc, rewriter.getI32Type(),
+                                                   valueToPrint);
+    }
+
```
linuxlonelyeagle, line 192: For unsupported data types, you should return `failure`.

midend/lib/Conversion/LowerLinalgToGemmini/LowerLinalgToGemmini.cpp
```diff
@@ -75,6 +75,167 @@ class MatmulLowering : public OpRewritePattern<linalg::MatmulOp> {
...
+    DenseIntElementsAttr stridesAttr = convOp.getStridesAttr();
+
+    size_t dilations = 1;
+    size_t strides = 1;
+    if (dilationsAttr)
+      dilations = (*dilationsAttr.begin()).getLimitedValue();
+    if (stridesAttr)
+      strides = (*stridesAttr.begin()).getLimitedValue();
+
+    if (inputShape[1] != inputShape[2]) // h, w
```
linuxlonelyeagle, line 110: The comment should be more explicit and should be placed above the if statement.

linuxlonelyeagle:
Ping @shirohasuki I don't seem to have approval authority. If no one approves this PR (CI cannot run, please @zhanghb97 in time.

shirohasuki:
Thank you so much for the REVIEW! I'll modify it and @zhanghb97 for CI.

linuxlonelyeagle:
Print Ops in Gemmini has always been coupled.Its logic should actually be `memref` -> `affine` -> `scf` -> `llvm`,In fact, Gemmini is just `gemmini` -> `llvm`,So maybe you can separate the printOp patterns into a new pass.You can see that the `populate pattern functions` from the upstream are introduced in the lower gemmini process.

shirohasuki:
I think it's right, Separate the printOp pattern into a decoupled pass and it will be like this?
```
populateGemminiPrintLegalizeForLLVMExportPatterns()
populateGemminiLegalizeForLLVMExportPatterns()
then, affine -> scf -> llvm................
```

shirohasuki:
> > I think it's right, Separate the printOp pattern into a decoupled pass and it will be like this?
> > ```
> > populateGemminiPrintLegalizeForLLVMExportPatterns()
> > populateGemminiLegalizeForLLVMExportPatterns()
> > then, affine -> scf -> llvm................
> > ```
> 
> `populateGemminiLegalizeForLLVMExportPatterns()` should only include patterns that are `lower to the gemmini instructions`. And `populateGemminiPrintLegalizeForLLVMExportPatterns()` should only include patterns for `print ops`, which lower print ops to MLIR upstream ops.and then use the upstream pass, which should be lower to llvm dialect.
> 
> ```
> gemmini.print => affine/scf
> // then use spsteam pass
> affine/scf => llvm dialect
> ```

I got it!

zhanghb97:
Please rebase the code. Let's check if it passes the CI tests after rebase.

shirohasuki:
I sync from the remote repo. Maybe we can try CI again?

linuxlonelyeagle:
You can see how much code you've changed in the upper right corner.Looks like you introduced someone else's PR.You should go back to the very first commit and then use git rebase your code to eliminate code conflicts.Or pulling the contents of the main branch should do the trick.

shirohasuki:
OKOK.

PR #464, opened 2025-02-19
[Test] Add test cases for python binding

JuniMay:
This PR adds several test cases for buddy-mlir python binding.

- Test cases for DIP and DAP dialects are derived from MLIR files under tests/Dialect/.
- Scalable vector creation in python binding is not yet supported by the current llvm version. I can rewrite the uses of parse in RVV test cases into step-by-step construction of operations once the llvm is updated.


zhanghb97:
Thanks!

PR #471, opened 2025-03-03
[midend] Support scalable vector for matmul-transpos-b vectorization

JuniMay:
This PR adds scalable vector support to matmul-transpose-b vectorization. `vec-size` is renamed to `vf` (as vectorization factor), and `scalable` option is added to enable scalable vector.

The step is now computed with `vscale * vf` and a symbol is added into the affine maps.

The performances are just the same compared to static vector method on musepi:

```
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
benchMatmulTransposeB/matmul             333 ms          333 ms            2
benchMatmulTransposeB/matmul_opt0        108 ms          108 ms            6
benchMatmulTransposeB/matmul_opt1        109 ms          109 ms            6
------------------------------------------------
MatmulTransposeB-opt0 PASS
MatmulTransposeB-opt1 PASS
------------------------------------------------
```

where opt0 is static vector with `vf=32` and `-mllvm --riscv-v-vector-bits-min=256` appended to clang args, and opt1 is scalable with `vf=8` (without explicit vlen arguments).

The tail processing is kept the same for now. I will resolve conflicts when it is modified (if there are any).

### DIFF
--- a/midend/lib/Conversion/MatMulOptimization/MatMulTransposeBVec.cpp
+++ b/midend/lib/Conversion/MatMulOptimization/MatMulTransposeBVec.cpp
@@ -117,16 +117,29 @@ public:
                   ValueRange itrArgs) {
                 Value acc = itrArgs[0];
 
-                AffineExpr a, b, c;
-                bindDims(ctx, a, b, c);
-                AffineMap AVectorMap = AffineMap::get(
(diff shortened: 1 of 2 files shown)


zhanghb97 (approved):
Thanks!

PR #476, opened 2025-03-05
increase the max-line-length

LIUQyou:
increase the max-line-length from 88 to 120


WuXintong123:
This pull request passes the test.

PR #479, opened 2025-03-13
[midend] Support arbitrary float type of zero constants

JuniMay:
The `insertZeroConstantOp` utility is used in matmul vectorization passes, but it currently only supports `f32` and `f64`. This PR adds support for other float types.


zhanghb97 (approved):
Thanks!

PR #515, opened 2025-07-14
feat: add torch dynamo and triton based backend inductor docs and example.

jackfiled:
- Add a doc which describes how PyTorch 2.0 use the triton to accelerate kernel execution on GPU.
- Add an example that shows the process a `ResNet 18` model run on CPU and run on GPU by the torch dynamo.


zhanghb97 (approved):
Thanks!

PR #532, opened 2025-08-22
[BUG]Add missing slash in CMakeLists.txt

x12301450:
In `examples/BuddyStableDiffusion/CMakeLists.txt` line 3, the cmake scripts is missing the required slash:

`${CMAKE_CURRENT_BINARY_DIR}arg1_text_encoder.data`

I think this should be fixed. Lookforward to your feedback!


zhanghb97 (approved):
Thanks!

PR #546, opened 2025-09-08
[VIR] Lower affine.for and memref.load in the region of vir.set_vl

CBalaa:
Lower `affine.for` inside the `set_vl` region.
Changes File `midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp`:
* Add function `findAnchorTypeRecursive` to recursively search for anchor type within a given Region and its nested.
* Add function `lowerBlock` to recursively lower operations within a given block.
* Change function `matchAndRewrite`, move if-else in `affine-for` builder to `lowerBlock`.


midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
   bool verbose;
   bool useCustomVecWid;
 
+  /// @brief Recursively search for anchor types within a Region and its nested
```
linuxlonelyeagle: Here, you can directly describe the function's purpose without explaining the parameters' meanings, to ensure consistent coding style throughout the project.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+  /// Regions
+  /// @param region The region to search for
+  /// @param anchorType Used to store references of found anchor types
+  /// @return If the anchor type is found, return true
+  bool findAnchorTypeRecursive(Region &region, Type &anchorType) const {
```
linuxlonelyeagle: Could you modify the implementation of this function to return a type? If no suitable type is found, simply return null. That should be better.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+    if (anchorType)
+      return true; // If found, exit early
```
linuxlonelyeagle: The comment is written above the if statement, but I actually don't think a comment is necessary here.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+
+    for (Block &block : region) {
+      for (Operation &op : block) {
+        // Check operands and results to find dynamic vector types
+        auto checkValue = [&](Value value) {
```
linuxlonelyeagle: Can anonymous functions be written outside a for loop?

linuxlonelyeagle: I don't think returning a bool type from this anonymous function is the simplest implementation. Perhaps you could return a type, as mentioned above.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+                  dyn_cast<vir::DynamicVectorType>(value.getType())) {
+            if (!dynVecType.getScalingFactor()) {
+              anchorType = dynVecType.getElementType();
+              return true;
+            }
+          }
+          return false;
+        };
+
+        for (Value operand : op.getOperands())
```
linuxlonelyeagle: If you use this checking method, a value may be checked multiple times.Let's consider this: Can I check the parameters of a block and then only check the result of the value?
``` mlir
func.func @foo(%a, %b) {
  %c = use(%a, %b) 
   %d = use(%a, %c)
    xxxx
}
```

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+        for (Region &innerRegion : op.getRegions()) {
+          if (findAnchorTypeRecursive(innerRegion, anchorType))
+            return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  /// @brief Recursively reducing operations within a block
```
linuxlonelyeagle: as above.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+  /// @param loc The current location
+  /// @param block  Block to lower
+  /// @param symbolTable Symbol table, used to map old values to new values
+  /// @param mainLoopIV Inductive variable (IV) of the main quantization loop
+  /// @param targetVectorType The target vector type can be null ptr for tail
+  /// loops
+  /// @param isTailLoop Flag, indicating whether the tail loop is currently
+  /// being processed (generating scalar code)
+  void lowerBlock(OpBuilder &builder, Location loc, Block *block,
+                  DenseMap<Value, Value> &symbolTable, Value mainLoopIV,
```
linuxlonelyeagle: Can we rename the variable symbolTable? Its name is misleading, as it suggests it's a symbol table class in MLIR.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+          .Case<vir::LoadOp>([&](vir::LoadOp op) {
+            Value base = op.getBase();
+            // the lastest index is dynamic and global
+            Value lastIndex = op.getIndices().back();
+            Value baseOffset =
+                builder.create<arith::AddIOp>(loc, mainLoopIV, lastIndex);
+            auto sourceIndices = op.getIndices();
+            SmallVector<Value> destIndices{};
+            // collect indices for load op
+            for (size_t i = 0; i < sourceIndices.size() - 1; i++) {
```
linuxlonelyeagle: for (size_t i = 0, e = sourceIndices.size() - 1; i < e; ++i) { or for (auto it : sourceIndices)

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -271,66 +273,277 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+            }
+            Value base = op.getBase();
+            // the lastest index is dynamic and global
+            Value lastIndex = op.getIndices().back();
+            Value baseOffset =
+                builder.create<arith::AddIOp>(loc, mainLoopIV, lastIndex);
+            auto sourceIndices = op.getIndices();
+            SmallVector<Value> destIndices{};
+            // collect indices for load op
+            for (size_t i = 0; i < sourceIndices.size() - 1; i++) {
```
linuxlonelyeagle: as above

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
zhanghb97: I think the term _symbol table_ is already accurate. If we really need to rename it, we could add a `vir` prefix, for example `virSymbolTable`.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: Sure, I do think so.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: I will simplify it.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: I change the signature of  function `findAnchorTypeRecursive` to `Type findAnchorTypeRecursive(Region &region)`, so this if statement is removed.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: > Can anonymous functions be written outside a for loop?

I move it to the outside of the loop.

CBalaa: > I don't think returning a bool type from this anonymous function is the simplest implementation. Perhaps you could return a type, as mentioned above.

I will change the funtion's signature.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: This is better. I take this: `for (size_t i = 0, e = sourceIndices.size() - 1; i < e; ++i) { `.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: Already changed.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: It is used for mapping old values in `block` (the function parameter), to the new values (built by `builder`),  I will add a vir prefix.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
CBalaa: You are right, dynamic vector types only exist in `set_vl` region, so only checking the result of the operation in region is better.

zhanghb97 (approved):

PR #547, opened 2025-09-12
[Examples]: Add Intel AMX BF16 matrix multiplication.

Jiajun-Ji:
Implements AMX tile operations with system permission handling and comprehensive benchmarking. Includes optimized linalg baseline, AOT compilation chain. Achieves ~2× speedup on Sapphire Rapids 8488C (512×2048×1024 matrices).
FileCheck testing is not supported because AMX tile operations require system-level permissions and AOT compilation; JIT-based testing frameworks like FileCheck cannot initialize AMX state or handle required syscalls.

### DIFF
--- a/examples/BuddyMatmul/amx-wrapper.c
+++ b/examples/BuddyMatmul/amx-wrapper.c
@@ -1,11 +1,32 @@
+//===- amx-wrapper.c - AMX Permission and MLIR Entry Wrapper --------------===//
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


examples/BuddyMatmul/amx-wrapper.c
zhanghb97, line 1: Please use `clang-format` to format this code and add the licence header.

Jiajun-Ji: Hi, I have format the .c code and add the licence header.

PR #553, opened 2025-09-24
[Frontend] Lower attention op to FlashAttention vectorized kernel.

GuoningHuang:
1. Implement FlashAttention kernel in MLIR using affine loops + vector ops.

2. Replace the existing Attention → TOSA lowering path with Attention → FlashAttention kernel in MLIR using affine loops + vector ops.


GuoningHuang:
This is a longer prompt test result：
Before：image
After：屏幕截图 2025-11-16 155404

PR #566, opened 2025-09-30
[midend] Fix initialization bug in matmul vectorization pass.

CBalaa:
Current matmul vectorization pass matches linalg.matmul and generates code that computes A@B, which is different from  linalg.matmul(A@B+C). 

This PR fix this bug, and now matmul vectorization pass generates code computing A@B+C.


CBalaa:
> Can you write an integration test file to test it?

Already added. And I add tail process for unroll dimension, so it can works for any size of matmul.

PR #587, opened 2025-10-20
[Frontend] Implement Online LayerNorm with Frontend Fusion and Lowering Pass Support

GuoningHuang:
This PR introduces an optimized implementation of Online LayerNorm 
It provides both frontend operator fusion and the corresponding lowering pass to support end-to-end execution.

1. Added an Online LayerNorm kernel based on the Welford algorithm for incremental mean/variance computation,achive 50x performance speed up compare with origin TOSA version.
2. Implemented frontend LayerNorm fusion, combining  operations (reduce_sum, rsqrt, mul, add, etc.) into a single fused operator.
3. Added lowering pass from fused LayerNorm to optimized vectorized IR.
4. Verified correctness on test_layernorm.py.
5. Can be directly integrated into the existing Deepseek R1 inference pipeline.
image

### DIFF
--- a/tests/Python/test_layernorm.py
+++ b/tests/Python/test_layernorm.py
@@ -1,0 +1,69 @@
+# RUN: %PYTHON %s 2>&1 | FileCheck %s
+
+import torch
+from torch._inductor.decomposition import decompositions as inductor_decomp
+
+from buddy.compiler.frontend import DynamoCompiler
+from buddy.compiler.ops import tosa
+
+from buddy.compiler.graph.transform import simply_fuse,apply_classic_fusion
+
+
+def kernel(x, gamma):
+    eps = 1e-5
+    x_pow = torch.pow(x, 2)
+    var = torch.mean(x_pow, dim=-1, keepdim=True)
+    inv_std = torch.rsqrt(var + eps)
+    norm = x * inv_std
+    out = gamma * norm  
+    return out
+
+x = torch.randn(1, 40, 1536)
+gamma = torch.randn(1536)
+
+# Initialize the dynamo compiler.
+dynamo_compiler = DynamoCompiler(
+    primary_registry=tosa.ops_registry,
+    aot_autograd_decomposition=inductor_decomp,
+)
+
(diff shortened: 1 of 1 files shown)


GuoningHuang:
> What about the performance of E2E integration? Has the issue of performance decline in E2E been resolved? Now we have the KV Cache support. You can use the latest version to integrate and test.

OK，the optimization on this operator still faces a performance decline. I think the affine + vector optimization strategy may not work, I'm working on another optimization strategy.

CBalaa:
I implemented flash attention using mlir [here](https://github.com/CBalaa/flash-attention-cpu).  You can refer to the code here to add flash attention to the buddy frontend.

GuoningHuang:
> @GuoningHuang I suggest not putting the fusion method in this PR and the existing fusion method (matmul transpose fusion) under the same option for now. The performance drop you mentioned in the [issue](https://github.com/buddy-compiler/buddy-mlir/issues/595) is actually due to the fact that the matmul transpose fusion hasn't integrated vectorization optimization yet (there's currently a memory layout issue being fixed).
> 
> I have tried your PR. If you disable the matmul transpose fusion, performance won't drop, but it also doesn't improve. This might be related to the fusion strategy issue that @CBalaa mentioned, as well as the fact that the fused code hasn't been optimized for vectorization and parallelization, so scalar execution doesn't show a significant performance advantage. I'm not sure where the 50x performance gain you mentioned is coming from.

OK, thank you for your suggestion!
The performance data I mentioned was tested using next-layernorm.mlir and next-layernorm-online.mlir.
I think the improvement might be because I used an incremental (online) RMSNorm to approximate the standard RMSNorm, which reduces both computation and memory access overhead.
However, I found this approximation may lead to some loss in inference accuracy, so I plan to explore other optimization approaches.

GuoningHuang:
> I implemented flash attention using mlir [here](https://github.com/CBalaa/flash-attention-cpu). You can refer to the code here to add flash attention to the buddy frontend.

Thank you! I will try it.

PR #590, opened 2025-10-21
Add traditional optimization pass to subgraph-f16

asdf1113:
Optimized the pipeline for deepseek f16, primarily adding traditional optimization passes: canonicalize sccp remove-dead-values 

Following these additions, the compilation time and maximum memory usage for buddy-opt decreased. Subsequently, the mlir-translate compilation time and maximum memory usage showed significant reductions: from 128.9s to 114.76s, and from 219MB to 189MB.

Although applying traditional optimization methods has little impact on the final program's performance, they can slightly reduce compilation memory and LLVM translation time.


zhanghb97:
Thanks!

PR #594, opened 2025-10-27
Completed the BLIS task

Yssx-g:
Refined the BLIS framework, verified the correctness of the code, and conducted performance comparisons.
(1):The following is a performance comparison of the matmul-vectorization algorithm and the custom Matmul-Blis algorithm on the server.
e21adba7be1448971c92b2e974909f3d
It can be observed that, under conditions of messy data or large data volumes, the matmul operation implemented using the BLIS framework demonstrates a significant performance improvement compared to the original matmul operation that directly performs vector expansion.
(2)The following are the results of integrating BLIS optimization into the end-to-end testing of DeepseekR1. The performance comparison results are on the right, and the correctness verification is on the left.
2489073537c1b816342327f2edf36201
These are all my commits.


zhanghb97 (approved):
Good progress!

PR #596, opened 2025-10-27
[Frontend] Achieve norm optimization by scf dialect and intergrate into E2E pass 

GuoningHuang:
This PR introduces an optimized implementation of Norm
It provides both frontend operator fusion and the corresponding lowering pass to support end-to-end execution.

Implemented frontend LayerNorm fusion, combining operations (reduce_sum, rsqrt, mul, add, etc.) into a single fused operator.
Added lowering pass from fused Norm to optimized vectorized IR.
Verified correctness on test_norm.py.
Can be directly integrated into the existing Deepseek R1 inference pipeline.
before opt:
image
after opt:
屏幕截图 2025-11-05 210652

### DIFF
--- a/examples/BuddyNext/next-norm-opt.mlir
+++ b/examples/BuddyNext/next-norm-opt.mlir
@@ -55,6 +55,7 @@ func.func @kernel(%arg0: tensor<1x1024x1536xf32>, %arg1: tensor<1536xf32>) -> te
     %zero = arith.constant 0.0 : f32
     %c0 = arith.constant 0 : index
     %c1 = arith.constant 1 : index
+    %c16 = arith.constant 16 : index
     %c1024 = arith.constant 1024 : index
     %c1536 = arith.constant 1536 : index
     %dim = arith.constant 1.536000e+03 : f32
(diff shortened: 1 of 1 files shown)


frontend/Python/ops/tosa.py
```diff
@@ -2047,6 +2050,91 @@ def broadcast_tensor(tensor, target_shape):
...
+    gamma_memref = bufferization.ToMemrefOp( ir.MemRefType.get(gamma.type.shape, f32), gamma)
+    out_memref = memref.AllocOp(ir.MemRefType.get(input_shape, f32), [], [])
+    # #Could choose to use scf.parallel to implement the loop_b or loop_i
+    # loop_b_parallel = scf.ParallelOp([], [c0.result], [dim1_index.result], [c1.result],[])
+    # blk = loop_b_parallel.region.blocks.append()
+    # with ir.InsertionPoint(blk):
+    #     b = blk.add_argument(ir.IndexType.get(), ir.Location.unknown())
+    loop_b = scf.ForOp(c0.result, dim1_index.result, c1.result, [])
+    with ir.InsertionPoint(loop_b.body):
+        b = loop_b.induction_variable
```
GuoningHuang, line 2115: #597

PR #618, opened 2025-11-12
[Bug] Fix conv2d example MLIR lowering pass pipeline

AkunCube:
Add `convert-arith-to-llvm` and `convert-cf-to-llvm` pass to conv2d example pipelines.


zhanghb97 (approved):
Thanks!

PR #640, opened 2025-12-09
[build][frontend] Add fallback for SPLITTING_SIZE

meshtag:
Add fallback split size when no SIMD feature or override is detected. This ensures that functionality is not broken on scalar-only targets.

CC: @zhanghb97


zhanghb97:
Thanks!

Issue #639, opened 2025-12-09
Refactor how frontend handles node datatypes


stomfaig:
While working on https://github.com/issues/mentioned?issue=buddy-compiler%7Cbuddy-mlir%7C605, I found that the following changes to how the frontend works internally would be really helpful:

1. **Refactor `Op` so that it contains only a single reference to its associated datatype**
  Currently there are multiple references within an `Op` node to its datatype, both through an argument, and also through `_tensor_meta`. This introduces non-indicated data dependencies, which have to be managed manually upon possible changes to dtype.
  *Implementation proposition:* Remove redundancy and use only one of these methods.
2. **Refactor how `Graph` handles `input` and `param` nodes.**  Currently, `Graph` has fields `input` and `fake_params`, which are completely independent of the `body`. This creates dependencies within `Graph` which are not clear from the data structure itself. This is also problematic, when we want to change the datatype of nodes (e.g. when quantising a network in passes, see https://github.com/issues/mentioned?issue=buddy-compiler%7Cbuddy-mlir%7C605), since we (a) need to change both the dtype of some nodes, and also of the entries in `input` and `fake_params` (b) the information about which `placeholder` nodes correspond to which `input` or `param` is lost once the graph has been constructed.
  *Implementation proposition:* Implement `input` and `param` via a stored index list to the corresponding nodes in `body` and potentially add getters to make accessing them clean.
  
These changes would allow to manipulate the dtype of nodes in passes without magic code that has to change a lot of coupled attributes.

We can discuss changes to implementation details.

R-Tars:
@stomfaig Thanks for the follow-up. Looking forward to your PR for point 2.

PR #641, opened 2025-12-13
[Frontend] Refactor how frontend handles node datatypes

stomfaig:
This PR implements a more direct reference system in `Graph` to different node types (e.g. `Input`, `Param` etc.). 

Some notes about the current version of the draft.

1. `GraphImporter` still also only receives `input` and `params` in `TensorMeta` (see ambiguity of this below). I've left it this way, because in `GraphImporter` we do not expect to change this again, so there is no reason to introduce more complication, though the notation will be improved to reflect this.

2. One point of friction here is that thorughout the frontend there seems to be two notions of `tensor_meta`, one being the type `TensorMeta`, the other being either a `dict[str, Any]`, the other being `dict[str, list[Any]]`. For now I left some convenience script in for the sake of this draft, but let's discuss what the general direction with this should be.

Closes buddy-compiler/buddy-mlir#639

### DIFF
--- a/frontend/Python/graph/graph_driver.py
+++ b/frontend/Python/graph/graph_driver.py
@@ -217,6 +217,12 @@ class GraphDriver:
             verbose=self._graph._verbose,
         )
 
+        # Adding placeholder operations from the original graph        
+        for op in self._graph.params:
+            main_graph.add_node(op, node_type=NodeType.FakeNode)
+        for op in self._graph.inputs:
+            main_graph.add_node(op, node_type=NodeType.InputNode)
+
         # Adding FuncOp nodes for each subgraph
         for subgraph_name in self._subgraphs.keys():
             func_node = FuncOp()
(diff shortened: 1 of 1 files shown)


R-Tars:
Thanks for pointing this out — I agree it is an important issue.

For now, I don’t think we should rush to unify `tensor_meta`, as doing so would likely require changes across a large amount of existing operator and frontend code. Given the scope of this PR, deferring that work seems reasonable.

In the longer term, I do think converging on `TensorMeta` as the unified representation would be preferable, but this can be revisited once the new reference system stabilizes.

R-Tars:
Thanks for looking into this. Just to add some context: a recent commit has caused most models to fail at runtime, so the frontend is currently in a rather unstable state. We are currently trying to address this issue. As a result, even if this specific issue is fixed, models may still not run correctly at the moment.

R-Tars:
It seems that this PR has not been updated for some time. I pulled the latest changes today and tested them locally against the current main branch. At the moment, the DeepSeek-R1 example still cannot run successfully, so the runtime issue does not appear to be fully resolved yet.

For this PR to be ready for merging, I think a minimum requirement is that the DeepSeek-R1 example can run correctly from start to finish. Currently, this is still not the case.

If you encounter any difficulties fixing the issue, please feel free to let me know — I am happy to help with debugging.

R-Tars:
Sorry, I apologize—this is indeed an oversight. In the current version, inference works for f32, but f16 and bf16 are not yet supported. I will fix this  as soon as possible.

stomfaig:
Ah, ok, no worries. I'll use f32 for now then.

R-Tars:
Thanks for the understanding. I’ve now merged a fix that updates the ClampOp invocation to match the new LLVM version. I just tested it locally, and f16 inference should work correctly now. Please feel free to give it another try when convenient.

R-Tars:
Thanks for the update, and glad to hear it now compiles locally.
I agree this placeholder ordering assumption is confusing and worth fixing — opening an issue for it sounds good.

PR #645, opened 2025-12-21
[frontend/tests] Add aten ops coverage

sdh1014:
## Motivation

PyTorch invokes operators through the dispatcher. While `torch._C._dispatch_get_all_op_names()` lists all registered ops (including many internal/backend-only entries), we focus on the Python-facing ATen surface under `torch.ops.aten`. In our environment, `torch.ops.aten` contains **1,920 public ATen operator overloads** (excluding leading-underscore ops, which are typically internal/unstable).

Our pipeline uses **decompositions** + **TorchDynamo** to capture operator calls and lower them into our graph IR. A previous commit covered a “core ATen IR” set, but under decomposition-driven capture we observed additional `torch.ops.aten` overloads that were not handled. This PR fills those gaps.

## Summary

- Under current decompositions, we lower **1,174 / 1,920** public `torch.ops.aten` overloads (**~61%**).
- Added batch conversion tests in `tests/Python/AtenOpsCoverage`.
- Added missing lowerings/mappings; explicitly skip unsupported ops with documented reasons (`RNG/rand`, `complex dtypes`, `backward`,`prims/quantized`,`sparse-only` and patterns not capturable by Dynamo in our current configuration such as certain `named-tensor` behaviors and `scalar/non-Tensor outputs`).
- Fixed lowering edge cases in `linalg/tosa` (indexing and `addmm` result handling).

## Note

This PR ensures IR generation succeeds for the covered ops; some lowerings may still need semantic refinement and will be validated as real workloads exercise the IR.


frontend/Python/frontend.py
```diff
@@ -136,10 +136,13 @@ def __init__(
             "neg.default": NegOp,
             "cat.default": CatOp,
             "bmm.default": BatchMatmulOp,
+            "div.default": DivOp,
             "div.Tensor": DivOp,
             "div.Tensor_mode": DivTensorModeOp,
+            "div.Tensor_mode": DivTensorModeOp,
```
zhanghb97: This looks like a duplicate addition (the operation below has some similar issues). What is the reason?

PR #651, opened 2025-12-24
[Frontend] Add quantisation pass

stomfaig:
This PR is implementing a new pass for quantising graphs. It includes a working initial version of a general quantization framework, as well as a few proof of concept quantization methods. Once it is landed, the rest of the quantization methods can be implemented independently / in parallel

Note: the code also contains the code from #641, so that it is runnable locally. This poc only supports simple models like below, and is intended as a demonstration of the approach. The changes _strictly associated with this pr_ are only in `frontend/Python/graph/transform/quantization`

Depends on #641, #644

### DIFF
--- a/frontend/Python/graph/graph_driver.py
+++ b/frontend/Python/graph/graph_driver.py
@@ -217,6 +217,12 @@ class GraphDriver:
             verbose=self._graph._verbose,
         )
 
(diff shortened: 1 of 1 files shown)


stomfaig:
cc: @R-Tars

PR #722, opened 2026-03-17
[Examples] add tfla op and tfla-optimized

Old-cpu:
### Description

This PR adds an optimized implementation of the Tiled Flash Linear Attention (TFLA) kernel based on the NeurIPS 2025 paper "Tiled Flash Linear Attention: More Efficient Linear RNN and xLSTM Kernels",demonstrating significant performance improvements over the baseline versions. The optimized TFLA kernel achieves approximately 21× speedup compared to the baseline TFLA implementation, and approximately 3× speedup compared to the fused GQA Attention kernel (next-gqa-attention-fusion.mlir).

### Hardware Configuration

  - CPU: Intel Xeon Silver 4114 @ 2.20GHz
  - Cores: 2 sockets × 10 cores = 20 physical cores (40 logical threads)
  - Architecture: x86_64 with AVX-512 support
  - OpenMP threads: 48

### Background

The TFLA kernel provides a fair comparison with Grouped Query Attention (GQA) under identical configurations:

- Batch size: 1
- Query heads: 12
- KV groups: 2 (each group serves 6 heads via GQA)
- Sequence length: 1 (single-token inference)
- Hidden dimension per head: 128
- KV cache sequence length: 1024 per groupt.

Expected Performance
im

- Describe how reviewers can verify and test this change. Include command-line instructions if applicable.
`make next-tfla-run` and `make next-tfla-optimized-run`

## Checklist

- [x] The code builds successfully
- [x] Existing tests pass
- [x] New tests are added where appropriate
- [x] Code follows the project coding style
- [x] Documentation is updated if needed

### DIFF
--- a/examples/BuddyNext/makefile
+++ b/examples/BuddyNext/makefile
@@ -2464,8 +2464,8 @@ next-tfla-run:
 		-one-shot-bufferize="bufferize-function-boundaries" \
 		-convert-linalg-to-affine-loops \
 		-affine-loop-fusion \
-		-lower-affine \
 		-convert-vector-to-scf \
+		-lower-affine \
 		-expand-strided-metadata \
 		-convert-scf-to-cf \
 		-convert-bufferization-to-memref \
--- a/examples/BuddyNext/makefile
+++ b/examples/BuddyNext/makefile
@@ -2447,3 +2447,80 @@ next-tosa-matmul-manual:
 	numactl --cpunodebind=0,1 --membind=0,1 \
 	taskset -c 0-47 \
 	./next-tosa-matmul-manual.out || true
+
+# ==============================================================================
+# TFLA: Tiled Flash Linear Attention
+# Based on NeurIPS 2025 paper "Tiled Flash Linear Attention: More Efficient Linear RNN and xLSTM Kernels"
+# ==============================================================================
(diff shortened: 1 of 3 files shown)


GuoningHuang:
1. For a fair comparison, I think we need to make the operator pass pipelines of both GQA attention and this TFLA the same. Here are the results after I made them consistent:
屏幕截图 2026-04-08 135106

2.  This kernel actually lacks numerical stability compared to GQA attention's softmax, which can cause precision degradation. So we need to use a numerically stable softmax. 
3. Also, I haven't figured out the difference between this attention and linear attention — this attention seems to just be the classic attention computation.

Old-cpu:
> 1. For a fair comparison, I think we need to make the operator pass pipelines of both GQA attention and this TFLA the same. Here are the results after I made them consistent:
> 
> 屏幕截图 2026-04-08 135106
> 2. This kernel actually lacks numerical stability compared to GQA attention's softmax, which can cause precision degradation. So we need to use a numerically stable softmax.
> 3. Also, I haven't figured out the difference between this attention and linear attention — this attention seems to just be the classic attention computation.

You're right, I did overlook a lot of things. I'll go and study it again. Thanks for your comment.🌹

PR #745, opened 2026-04-10
[VIR] Support Reduce 3D And Transpose

xbwcan:
## Summary
- Add support for linalg.reduce when rank=2, dim=0 and rank=3, dim=0.
- Add pattern to lower linalg.transpose to vector.transpose.
- Modified and added some test cases.

### DIFF
--- a/midend/lib/Conversion/LinalgToVIR/LinalgToVIRPass.cpp
+++ b/midend/lib/Conversion/LinalgToVIR/LinalgToVIRPass.cpp
@@ -382,6 +382,7 @@ static LogicalResult lowerReduceToScalarLoop(linalg::ReduceOp reduceOp,
   };
 
   // Case A: rank-1 -> rank-0, dimensions = [0].
+  // Reduce the full 1D buffer into a scalar accumulator.
   if (inTy.getRank() == 1 && outTy.getRank() == 0) {
     if (dims.size() != 1 || dims[0] != 0)
       return rewriter.notifyMatchFailure(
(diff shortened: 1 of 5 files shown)


xlinsist:
When a fully static linalg.transpose hits this rewrite, this PR now emit a vector.transpose instead of going through the existing VIR path, which can be seem as a temporary fix for triton2vir pipeline. Maybe `vir.transpose` can be added in the future.

midend/lib/Conversion/LinalgToVIR/LinalgToVIRPass.cpp
```diff
@@ -420,60 +421,240 @@ static LogicalResult lowerReduceToScalarLoop(linalg::ReduceOp reduceOp,
...
+          AffineMap::getPermutationMap(permutation, rewriter.getContext());
+      auto scope = rewriter.create<memref::AllocaScopeOp>(loc, TypeRange{});
+      Block *scopeBlock = rewriter.createBlock(&scope.getBodyRegion());
+      rewriter.setInsertionPointToStart(scopeBlock);
+
+      Value transposed = rewriter.create<memref::TransposeOp>(
+          loc, input, AffineMapAttr::get(permutationMap));
+      auto scratchTy = MemRefType::get({inStaticN, inStaticM}, elemTy);
+      Value scratch = rewriter.create<memref::AllocaOp>(loc, scratchTy);
+      rewriter.create<memref::CopyOp>(loc, transposed, scratch);
```
xlinsist, line 515: Are there ways for us to eliminate `memref.copy`? Since it may lead to bad performance.

xlinsist:
The missing example for lowering reduction on non-last dimensions of a high-dimensional array has been added, and `memref.alloca` has been replaced with `memref.alloc` for heap allocation — LGTM, thanks!

xlinsist:
Regarding the reduction on non-last dimensions of a high-dimensional array, the current approach allocates an auxiliary array of the same size as the input, performs an in-place transpose, applies the reduction primitive, and then writes the result directly back to the output array. The performance impact of this approach is not yet clear. This can be left as a follow-up issue to be addressed in a subsequent PR.

PR #747, opened 2026-04-10
[ci] Enable check-buddy on x64

trdthg:
## Summary

1. Fix ninja check-buddy buddy-lenet-run-test-cpu not found even if `DBUDDY_ENABLE_E2E_TESTS=OFF`
2. The current frontend hardcodes the search paths for the three libraries `libomp`, `libmlir_runner_util`, and `libmlir_c_runner_util`, causing one of the lit tests to fail when using LLVM build out-of-tree. The current solution is to add a new environment variable `LLVM_LIBS_DIR` to the frontend to discover LLVM libraries.


zhanghb97:
Please resolve the CI failures.

PR #751, opened 2026-04-13
[VIR] Fix the linalg.generic bug.

xbwcan:
…mension, and change the lowering target for non-unit-stride vir.vec

## Summary

- Force the last dimension of VIR to be generated as a dynamic dimension
- Set the lowering target for non-unit-stride vir.vec to vector.transfer_read/write


tests/Conversion/linalg-non-unit-stride-to-vir-to-vector.mlir
xlinsist, line 1: Move this test file into test/Conversion/LinalgToVIR.

PR #775, opened 2026-04-30
[ci] riscv64: GITHUB_ENV does not work in currently shell block

yuzibo:
Otherwise each cCI job will request to build LLVM modules.


yuzibo:
CC @trdthg

PR #788, opened 2026-05-10
[Midend] Support math.erf lowering via scalarization in VIRToVector

maochuanjun:
## Description
This PR adds `math.erf` lowering support in `VIRToVectorPass` to resolve the compilation failure of `tl.erf` in the Triton-RISCV backend (e.g.  `gelu_and_mul approximate="none"` from triton-riscv [#12](https://github.com/RuyiAI-Stack/triton-riscv/pull/12) and [#13](https://github.com/RuyiAI-Stack/triton-riscv/issues/13) ).

### `midend/lib/Conversion/LinalgToVIR/LinalgToVIRPass.cpp`
Add `math::ErfOp` to `isSupportedGenericBodyOp`.
### `midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp`
Add a module-level helper `getOrInsertErff()` and a `math::ErfOp` case.

## Implementation Approach
Used a scalarization strategy as a minimal fix:
1. Validates that the input is an `f32` scalar or a fixed-length `f32` vector.
2. Injects a symbol declaration for the standard C library `erff` function (`(f32) -> f32`).
3. For vectors, extracts elements sequentially, computes the `erff` value, and inserts the result back into a new vector allocation.

## Limitations
This is intentionally a narrow fix.
* **Type Restriction:** Currently restricted to `f32` to prevent implicit precision loss or incorrect libc linking.
* **Scalability:** Does not support scalable vectors (`assert` enforced) as this is designed as a direct, minimal fallback.

### DIFF
--- a/midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
+++ b/midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
@@ -131,6 +131,21 @@ static void buildScalarizedI1VectorStore(OpBuilder &builder, Location loc,
   }
 }
 
+static func::FuncOp getOrInsertErff(OpBuilder &builder, ModuleOp moduleOp) {
+  if (auto existing = moduleOp.lookupSymbol<func::FuncOp>("erff"))
+    return existing;
+
+  OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointToStart(moduleOp.getBody());
+
+  auto f32Ty = builder.getF32Type();
+  auto funcType = builder.getFunctionType({f32Ty}, {f32Ty});
+  auto funcOp =
+      builder.create<func::FuncOp>(builder.getUnknownLoc(), "erff", funcType);
+  funcOp.setPrivate();
+  return funcOp;
+}
+
 VectorType vectorTypeHelper(Region &srcRegion, Type anchorType,
                             bool useScalable, bool verbose) {
 #define VERBOSE_ANALYSIS_INFO(info)                                            \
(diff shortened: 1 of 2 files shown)


xlinsist:
Scalar handling of erf is acceptable since its vectorization is not fully supported. Since the optimization approach for math.erf is not exactly the same as math.exp, I suggest adding separate test cases for math.erf, then resolve the conflicts and fix the CI, and we are ready to merge it.

midend/lib/Conversion/VIRToVector/VIRToVectorPass.cpp
```diff
@@ -1181,6 +1196,65 @@ class VIRSetVLLowering : public ConvertOpToLLVMPattern<vir::SetVLOp> {
...
+                return isa<FloatType>(vt.getElementType()) &&
+                       cast<FloatType>(vt.getElementType()).isF32();
+              return false;
+            };
+            if (!isF32Value(in)) {
+              op.emitError(
+                  "math.erf lowering in triton-riscv only supports f32 scalar "
+                  "or vector inputs");
+              return;
+            }
```
xlinsist, line 1264: Since the VIRToVector path may not be exclusively dependent on triton-riscv in the future, it's better not to mention triton-riscv here. Also, there's no need to emphasize "supports scalar or vector inputs" to users, just mention what kind of data types are supported.

maochuanjun: already removed relative mention

maochuanjun:
Thanks for the review!

I have updated the branch and resolved the conflicts, so everything should be good to go now. I also added a separate test `tests/Conversion/math-erf-lowering.mlir`, which passes successfully on my locally.

Looking forward to your feedback,.Hope to be merged smoothly!

PR #813, opened 2026-06-24
[Conversion][BOSCAME] Add LowerLinalgToBOSCAME pass and update LLVM submodule

Yranger:
…ubmodule

## Summary

- This pull request introduces the `LowerLinalgToBOSCAME` conversion pass. It maps generic `linalg` operations (such as matrix multiplications) to the hardware-specific BOSC AME dialect.
- It also bumps and updates the `llvm` submodule pointer to the latest backend commit which includes the decoupled BOSC AME physical registers and intrinsic definitions.
- Reviewers can verify the lowering pipeline using the provided `buddy-opt` pass and the test cases under `examples/BOSCAMEDialect/`.

## Checklist

- [x ] The code builds successfully
- [x ] Existing tests pass
- [x ] New tests are added where appropriate
- [x ] Code follows the project coding style
- [x ] Documentation is updated if needed


backend/include/llvm/IR/IntrinsicsRISCVBuddyBOSCExt.td
trdthg, line 1: Since you've already added these to LLVM, you don't need to add these two files here again, right?
We could actually delete the entire backend folder, but I'd like to wait a little longer.

examples/BOSCAMEDialect/makefile
```diff
@@ -2,7 +2,7 @@
 
 BUDDY_OPT := ../../build/bin/buddy-opt
 BUDDY_TRANSLATE := ../../build/bin/buddy-translate
-BUDDY_LLC := ../../build/bin/buddy-llc
+BUDDY_LLC := ../../llvm/build/bin/llc
```
trdthg: I've already added buddy-llc to the dependencies of `check-buddy`, so the code here shouldn't need to be modified?

backend/include/llvm/IR/IntrinsicsRISCVBuddyBOSCExt.td
Yranger: Yes, I think the backend can be deleted, waiting is fine too.

examples/BOSCAMEDialect/makefile
Yranger: But if I don't change it, I won't be able to find llc during compilation, so I'm using the llc from the LLVM backend.

trdthg: I just tested it again, `buddy-llc` is now a dependency of `check-buddy`. After you run `check-buddy`, `buddy-llc` should appear in the `build/bin` folder. This is my runtime log.

image

trdthg: Or maybe you can add buddy-llc to this list

https://github.com/buddy-compiler/buddy-mlir/blob/04a4d0c8e9a1f3167d460007c2b1cb40fb2874fd/examples/CMakeLists.txt#L87-L92

Yranger: I've tested it again just now, and it really works for checking. I don't know why it didn't work before, but now both buddy-llc under buddy and llc under llvm can be executed.

Yranger: Then I'll keep it as it is.

PR #826, opened 2026-07-02
Document model integration baseline

GuoningHuang:
## Summary

This PR adds tools/buddy-codegen/MODEL_INTEGRATION_BASELINE.md, a machine-readable specification document intended primarily to be read by AI coding agents (and human contributors) when integrating a new packaged model into buddy-mlir.
The document defines a single, canonical integration path so that every future model follows the same shape, uses the same shared tooling, and produces the same predictable artifacts — instead of each model growing its own private build scripts and ad-hoc conventions.


GuoningHuang:
close

PR #869, opened 2026-08-09
[Conversion][GPU] Lower strided memref.copy for GPU

shirohasuki:
## Summary

NVVM cannot lower gpu.memcpy on non-identity layouts (e.g. RoPE half-dim subviews). Convert those copies to linalg.generic first and reject leftover strided memref.copy in convert-memcpy-to-gpu.

## Checklist

- [x] The code builds successfully
- [ ] Existing tests pass
- [ ] New tests are added where appropriate
- [x] Code follows the project coding style
- [ ] Documentation is updated if needed

### DIFF
--- a/midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
+++ b/midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
@@ -166,12 +166,25 @@ void ConvertMemcpyToGPUPass::runOnOperation() {
     else if (auto copyOp = dyn_cast<memref::CopyOp>(nestedOp)) {
       auto src = copyOp.getOperand(0);
       auto dst = copyOp.getOperand(1);
+      auto srcType = dyn_cast<MemRefType>(src.getType());
+      auto dstType = dyn_cast<MemRefType>(dst.getType());
+      if (!srcType || !dstType) {
+        copyOp.emitOpError("expected memref operands");
+        signalPassFailure();
+        return WalkResult::interrupt();
+      }
+      if (!srcType.getLayout().isIdentity() ||
+          !dstType.getLayout().isIdentity()) {
+        copyOp.emitOpError("strided memref.copy must be converted by "
+                           "convert-strided-memref-copy-to-linalg before "
+                           "convert-memcpy-to-gpu");
+        signalPassFailure();
+        return WalkResult::interrupt();
+      }
       // Notice: GPU.memcpy has a different src dst order
       builder.setInsertionPointAfter(copyOp);
-      auto gpuMemcpyOp = gpu::MemcpyOp::create(
-          builder, copyOp->getLoc(), TypeRange(), ValueRange(), dst, src);
-      src.replaceAllUsesWith(gpuMemcpyOp->getResult(1));
-      dst.replaceAllUsesWith(gpuMemcpyOp->getResult(0));
+      gpu::MemcpyOp::create(builder, copyOp->getLoc(), TypeRange(),
+                            ValueRange(), dst, src);
       copyOp->erase();
     }
     // Allocate space on GPU and copy global memrefs to GPU, needs deallocation
--- a/midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
+++ b/midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
@@ -18,33 +18,23 @@
 //
 //===---------------------------------------------------------------------===//
 
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/IR/AffineMap.h"
-#include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinTypes.h"
-#include "mlir/IR/OperationSupport.h"
-#include "mlir/IR/TypeRange.h"
-#include "mlir/IR/Types.h"
-#include "mlir/IR/ValueRange.h"
+#include "mlir/IR/PatternMatch.h"
 #include "mlir/IR/Visitors.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
+#include "mlir/Pass/Pass.h"
 #include "mlir/Support/LLVM.h"
 #include "llvm/ADT/SmallVector.h"
-#include "llvm/Support/Casting.h"
-#include "llvm/Support/ErrorHandling.h"
-#include <mlir/Dialect/Affine/IR/AffineOps.h>
-#include <mlir/Dialect/Func/IR/FuncOps.h>
-#include <mlir/Dialect/Linalg/Transforms/Transforms.h>
-#include <mlir/IR/Dialect.h>
-#include <mlir/IR/Operation.h>
-#include <mlir/IR/TypeUtilities.h>
-#include <mlir/IR/Value.h>
-#include <mlir/Pass/Pass.h>
 
 #include <vector>
 
 using namespace mlir;
-using namespace vector;
 
 //===----------------------------------------------------------------------===//
 // ConvertMemcpyToGPUPass
@@ -53,7 +43,8 @@ using namespace vector;
 namespace {
 
 class ConvertMemcpyToGPUPass
-    : public PassWrapper<ConvertMemcpyToGPUPass, OperationPass<func::FuncOp>> {
+    : public PassWrapper<ConvertMemcpyToGPUPass,
+                         InterfacePass<FunctionOpInterface>> {
 public:
   MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(ConvertMemcpyToGPUPass)
   StringRef getArgument() const final { return "convert-memcpy-to-gpu"; }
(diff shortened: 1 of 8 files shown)


midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
```diff
@@ -166,12 +166,25 @@ void ConvertMemcpyToGPUPass::runOnOperation() {
     else if (auto copyOp = dyn_cast<memref::CopyOp>(nestedOp)) {
       auto src = copyOp.getOperand(0);
       auto dst = copyOp.getOperand(1);
+      auto srcType = dyn_cast<MemRefType>(src.getType());
+      auto dstType = dyn_cast<MemRefType>(dst.getType());
+      if (!srcType || !dstType) {
+        copyOp.emitOpError("expected memref operands");
+        signalPassFailure();
+        return WalkResult::interrupt();
```
linuxlonelyeagle, line 164: Here are a few details: you can directly return WalkResult::interrupt();, and then check the return value of the walk method to determine whether signalPassFailure(); needs to be called.

midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
```diff
@@ -166,12 +166,25 @@ void ConvertMemcpyToGPUPass::runOnOperation() {
...
+      }
       // Notice: GPU.memcpy has a different src dst order
       builder.setInsertionPointAfter(copyOp);
-      auto gpuMemcpyOp = gpu::MemcpyOp::create(
-          builder, copyOp->getLoc(), TypeRange(), ValueRange(), dst, src);
-      src.replaceAllUsesWith(gpuMemcpyOp->getResult(1));
-      dst.replaceAllUsesWith(gpuMemcpyOp->getResult(0));
+      gpu::MemcpyOp::create(builder, copyOp->getLoc(), TypeRange(),
+                            ValueRange(), dst, src);
```
linuxlonelyeagle, line 176: Another detail is that this pass seems to create ops inside the walk method, and the logic inside walk is really quite complex. If possible, I think you might want to refactor this pass.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Pass/Pass.h"
+
+using namespace mlir;
+
+namespace {
+
+class ConvertStridedMemrefCopyToLinalgPass
+    : public PassWrapper<ConvertStridedMemrefCopyToLinalgPass,
+                         OperationPass<func::FuncOp>> {
```
linuxlonelyeagle: Maybe you can use functionOpInterface, rather than func::func.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
```diff
@@ -0,0 +1,83 @@
...
+    func::FuncOp funcOp = getOperation();
+    SmallVector<memref::CopyOp> copies;
+    funcOp.walk([&](memref::CopyOp op) { copies.push_back(op); });
+
+    for (memref::CopyOp copyOp : copies) {
+      Value src = copyOp.getSource();
+      Value dst = copyOp.getTarget();
+      if (src == dst) {
+        copyOp.erase();
+        continue;
```
linuxlonelyeagle: There's no need to write this logic—I believe a pattern like this already exists in canonicalization.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      auto srcType = dyn_cast<MemRefType>(src.getType());
+      auto dstType = dyn_cast<MemRefType>(dst.getType());
+      if (!srcType || !dstType) {
+        copyOp.emitOpError("expected memref operands");
+        signalPassFailure();
+        return;
```
linuxlonelyeagle: can you `return signalPassFailure()`?

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
```diff
@@ -0,0 +1,83 @@
...
+      if (srcType.getRank() != dstType.getRank()) {
+        copyOp.emitOpError("rank mismatch between source and target");
+        signalPassFailure();
+        return;
+      }
+
+      OpBuilder builder(copyOp);
+      linalg::makeMemRefCopyOp(builder, copyOp.getLoc(), src, dst);
+      copyOp.erase();
+    }
```
linuxlonelyeagle, line 79: * create the builder out of the for loop.
* use builder.eraseOp() method.

linuxlonelyeagle (commented):
I left some feedback here.

shirohasuki:
> I left some feedback here.

Thank you for your comments. I will make the revisions!

midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
shirohasuki: Thanks. I have split them into a separate entry
```
  if (walkResult.wasInterrupted())
    return signalPassFailure();
```

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
shirohasuki: I try to do that.

It can't auto-nest like this
092a6bdeac4ed50ab48c3f2d3d8f6112

So, it needs to be integrated into `builtin.module(func.func(`. I'm not sure if this is worth it.

To maintain consistency with `convert-memcpy-to-gpu`, I haven’t made this change for the time being.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
shirohasuki: Got it!

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
shirohasuki: fixed

midend/lib/Conversion/MLIRGPU/ConvertMemcpyToGPU.cpp
shirohasuki: I agree the walk in ConvertMemcpyToGPU is doing too much (alloc / copy / get_global / return) and mutating IR while walking. I’d prefer another separate PR for that.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
shirohasuki: - fixed
- OpBuilder has no eraseOp here; So I keep copyOp.erase() to stay consistent with ConvertMemcpyToGPU

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
linuxlonelyeagle: > So, it needs to be integrated into builtin.module(func.func(. I'm not sure if this is worth it.

If you are writing a Pass that you want to run on func.func rather than builtin.module, you should use FunctionOpInterface. Using syntax like builtin.module(func.func(...)) is totally Common.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
linuxlonelyeagle: You can use `IRRewriter` replace `OpBuilder` and use `erase` method.

shirohasuki: Fixed in the latest commit.

midend/lib/Conversion/MLIRGPU/ConvertStridedMemrefCopyToLinalg.cpp
shirohasuki: Refactor in the latest commit.

shirohasuki:
Thanks for your time @linuxlonelyeagle . I've updated both passes to InterfacePass<FunctionOpInterface>, and switched to IRRewriter + eraseOp, in the latest commit.

tests/Conversion/convert-strided-memref-copy-to-linalg.mlir
```diff
@@ -0,0 +1,44 @@
...
+    memref.copy %arg0, %arg1
+      : memref<4x8xf32, strided<[?, ?], offset: ?>> to memref<4x8xf32>
+    return
+  }
+
+  // CHECK-LABEL: func.func @identity_copy
+  // CHECK: memref.copy
+  // CHECK-NOT: linalg.generic
+  func.func @identity_copy(%arg0: memref<4x8xf32>, %arg1: memref<4x8xf32>) {
+    memref.copy %arg0, %arg1 : memref<4x8xf32> to memref<4x8xf32>
```
linuxlonelyeagle: I think the test cases could be made much cleaner(It might be a bit nitpicky, but it's true).
for example
```
 // CHECK-LABEL: func @identity_copy
 //       CHECK: memref.copy
 //   CHECK-NOT: linalg.generic
```
Above and below.

tests/Conversion/convert-strided-memref-copy-to-linalg.mlir
```diff
@@ -0,0 +1,44 @@
+// RUN: buddy-opt -pass-pipeline="builtin.module(func.func(convert-strided-memref-copy-to-linalg))" %s | FileCheck %s
```
linuxlonelyeagle: And you can use `-split-input-file`, You can take a look at its usage.

linuxlonelyeagle (commented):
A few comments on the tests.

shirohasuki:
> A few comments on the tests.

Thanks for the review! I’ve cleaned up the FileCheck style and added -split-input-file.

linuxlonelyeagle (approved):
LGTM，thanks for your effort. Ping @zhanghb97 ,please do a final review.

zhanghb97:
@shirohasuki @linuxlonelyeagle Thanks!

PR #873, opened 2026-08-24
[frontend][midend] Add FP16 panel-packed decode matmul support

GuoningHuang:
## Summary

Extend the panel-packed decode matmul optimization introduced in #839 to FP16 and integrate it into both DeepSeek-R1 build paths.

- make `matmul-vectorization-decode-packed` explicitly support same-type floating-point A/B/C operands, including native `vector<32xf16>` loads, FMA, and stores
- add an FP16 packed-decode lit test and a ready-to-use `f16_packed_decode.json` buddy-codegen spec
- teach `examples/BuddyDeepSeekR1/import-deepseek-r1.py` to pack FP16 decode weights and emit a separate `arg0-decode.data`
- add the `buddy-deepseek-r1-packed-f16-run` example target while keeping prefill on the ordinary row-major FP16 weights
- make the example importer Python path explicit so the target builds in a clean configured tree

## Usage

```bash
cmake --build build --target buddy-deepseek-r1-packed-f16-run -j2
build/bin/buddy-deepseek-r1-packed-f16-run
```

For the buddy-codegen path, select:

```text
models/deepseek_r1/specs/f16_packed_decode.json
```

## Performance

DeepSeek-R1-Distill-Qwen-1.5B FP16, Intel Xeon Platinum 8488C, 48 CPUs (`0-47`), NUMA nodes `0,1`, greedy decode, with identical output across the two runs (111 generated tokens):

| Configuration | Decode | Prefill | Total |
| --- | ---: | ---: | ---: |
| Plain FP16 | 25.2014 tokens/s | 8.93988 s | 13.3444 s |
| Packed FP16 | 59.9245 tokens/s | 8.79446 s | 10.6468 s |

This is a **2.38x decode speedup**. End-to-end time decreases by **20.2%**, while prefill remains effectively unchanged.

The standalone example path also completed end-to-end generation successfully at 55.31 tokens/s decode.

### Benchmark screenshots

Plain FP16:
image

Packed FP16:
image

## Testing

- rebuilt `buddy-opt`
- passed the explicit-shape and pack-all packed decode FileCheck tests
- verified FP16 lowering from linalg/vector MLIR through LLVM IR to an x86-64 ELF object
- generated the FP16 buddy-codegen configuration and verified its FP16 weights and decode weight entry
- built `buddy-deepseek-r1-packed-f16-run` from the example path
- ran plain and packed FP16 `.rax` models end to end and confirmed identical generated tokens
- ran the new FP16 packed example executable end to end

### DIFF
--- a/docs/DecodeWeightPacking.md
+++ b/docs/DecodeWeightPacking.md
@@ -74,6 +74,11 @@ Packing is off by default. It is enabled per spec, by setting
 }
 ```
 
+Ready-to-use DeepSeek-R1 specs are provided for both precisions:
+`models/deepseek_r1/specs/f32_packed_decode.json` and
+`models/deepseek_r1/specs/f16_packed_decode.json`. The FP16 path keeps the
+weights packed as FP16 values and emits native `vector<32xf16>` loads and FMA.
+
 `models/deepseek_r1/specs/f32_packed_decode.json` is the f32 DeepSeek R1 spec
 with packing enabled. Build it the usual way:
 
--- a/docs/DecodeWeightPacking.md
+++ b/docs/DecodeWeightPacking.md
@@ -109,8 +109,16 @@ Two restrictions are enforced at config time:
 ninja buddy-deepseek-r1-packed-run
 ```
 
+The FP16 example uses a separate target and output directory:
+
+```bash
+ninja buddy-deepseek-r1-packed-f16-run
+./bin/buddy-deepseek-r1-packed-f16-run
+```
+
 It imports with `--pack-decode-weights --pack-vector-size 32` and writes its
-artifacts to `packed_decode/` so they do not clobber the plain f32 outputs.
+artifacts to `packed_decode/` (FP32) or `packed_decode_f16/` (FP16), so they do
+not clobber the plain outputs.
 
 ## Adapting To Other Models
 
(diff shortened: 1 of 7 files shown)


zhanghb97:
Are there any experimental results for RISC-V devices?

GuoningHuang:
> Are there any experimental results for RISC-V devices?

Yes. The following results were measured on SG2044 using the FP16 inference path.

### Test configuration

- Platform: SG2044
- Precision: FP16
- `BUDDY_RISCV_ENABLE_ZFH_ZVFH=ON`
- Sequence length: 1024 (default)
- Packed vector size: 32
- Three runs per configuration; the median is reported
- Each run generated 63 decode tokens
- Regular and packed configurations produced identical output text

### Results

| Metric | Regular | Packed | Change |
|---|---:|---:|---:|
| Total time | 91.630 s | 84.732 s | 7.53% reduction, 1.081× speedup |
| Prefill throughput | 14.519 tokens/s | 14.289 tokens/s | -1.59% |
| Decode throughput | 3.126 tokens/s | 4.375 tokens/s | **+39.95%, 1.399× speedup** |

Packed weights improve median decode throughput by approximately 40%. Prefill uses the same plain-weight path in both configurations, so the 1.59% difference is within normal runtime variation.

PR #878, opened 2026-08-27
[frontend] Preserve template output order

asdf1113:
## Summary

This PR fixes the output ordering behavior of
`TemplatePartitionedGraphDriver` and makes model-specific runtime ABI
compatibility explicit.

Previously, when `output_remap` was not provided,
`construct_template_combined_main_graph()` reordered graph outputs by the
position of their producers in `graph.body`.

That behavior happened to match the historical output ordering of the legacy
`GraphDriver` for some models, but it also silently changed the semantic output
order represented by the graph `OutputOp`.

This PR:

- preserves the original graph output order by default in
  `TemplatePartitionedGraphDriver`;
- uses explicit `output_remap` for Qwen3-VL Vision to preserve its existing
  runtime ABI;
- uses explicit `output_remap` for Whisper to preserve its existing runtime
  ABI;
- adds a regression case whose graph output order is intentionally different
  from producer order.

## Background

The legacy `GraphDriver` and `TemplatePartitionedGraphDriver` construct the
combined forward graph differently.

The legacy `GraphDriver` is primarily organized around fused `op_groups`.
Outputs are naturally collected while traversing those groups, so the final
output order can follow producer/group traversal order rather than the operand
order of the original graph `OutputOp`.

The template-partitioned driver has a different structure. It materializes
reusable template subgraphs, invokes their concrete region instances in graph
order, and maintains mappings from original graph values to materialized SSA
values. Therefore, after computation is materialized, it can resolve public
outputs directly according to the original `OutputOp` order.

The previous implementation nevertheless sorted these resolved outputs again
by producer position when no explicit `output_remap` was supplied. This made a
legacy implementation detail part of the generic template driver's behavior.

### DIFF
--- a/frontend/Python/graph/partitioned_graph_driver.py
+++ b/frontend/Python/graph/partitioned_graph_driver.py
@@ -1747,11 +1747,6 @@ class TemplatePartitionedGraphDriver:
                     original_output, self._graph.node_table
                 )
             )
-        if output_remap is None:
-            body_positions = {
-                op: index for index, op in enumerate(self._graph.body)
-            }
-            final_outputs.sort(key=lambda value: body_positions[value.op])
         resolved_outputs = []
         for value in final_outputs:
             if value.op in self._plan.parameter_indices:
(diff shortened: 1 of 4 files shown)


zhanghb97:
The main branch has fixed the issue. Please rebase your code, and the CI should pass.

